PR #6, opened 2017-05-25
New Env Versions from latest Gym Version

karansaxena:
Updated environment versions according to the latest Gym version.

Refer PR: https://github.com/openai/gym/pull/584/files


siemanko:
Thanks! Resolved in recent commit.

PR #23, opened 2017-06-06
Namescopes

ngc92:
added some op names and name scopes to the Graph generation so that it is a bit easier to see 
what is going on in Tensorboard.


siemanko:
Hi!

Thanks for contributing this feature! Unfortunately we do not have capacity to properly review your change. At this point we only accept small changes such as typo and bug fixes.

Szymon

PR #30, opened 2017-06-12
Update setup.py

AdrianP-:
Fixed Error: ImportError: No module named 'cv2' 
That error appears when try to running train_pong.py


siemanko:
I think some people install this with conda, so forcing opencv-python might break their setup.

now-raymond:
What about people who _don't_ install this with conda? If conda is the only supported environment for baselines, you need to make this clear in the README.

PR #32, opened 2017-06-14
fix a numpy warning

ngc92:
don't exclude current epsiode when calculating 100 episode mean


siemanko:
Hi!

Thanks for contributing this feature! Unfortunately we do not have capacity to properly review and test your change.

Szymon

PR #33, opened 2017-06-16
Fix README since BreakOut pretrained model doesn't match the correct …

cxxgtxy:
…tensor shape. Therefore, Pong is used instead.


j8lp:
Agreed.  Or alternatively retrain breakout model against the latest Breakout environment.  It's

PR #34, opened 2017-06-17
Add rank-based prioritization

hamzamerzic:
In Schaul, Tom, et al. "Prioritized experience replay" two prioritized experience replay buffers are discussed: the proportional buffer, which was already contained in baselines, and the rank-based buffer, which is implemented as part of this PR.

### DIFF
--- a/baselines/deepq/__init__.py
+++ b/baselines/deepq/__init__.py
@@ -2,4 +2,4 @@ from baselines.deepq import models  # noqa
 from baselines.deepq.build_graph import build_act, build_train  # noqa
 
 from baselines.deepq.simple import learn, load  # noqa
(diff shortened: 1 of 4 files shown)


siemanko:
Hi!

Thanks for contributing this feature! Unfortunately we do not have capacity to properly review your change. At this point we only accept small changes such as typo and bug fixes.

Szymon

skroon:
I see a couple of contributions closed because of lack of capacity to review.  Perhaps it would be a good idea to make a fork of this repository where members of the community can pre-review and merge these features, so that they are better prepared for later integration into baselines if/when you do have capacity?

hamzamerzic:
@skroon I fully agree. I think there is definitely a need for that, and it is a waste to prevent people from contributing features already tested by the community.

PR #50, opened 2017-07-09
update tensorflow required version

GeeshangXu:
1. tensorflow change to GPU version
Keep current CPU version will cause failure to use GPU available. A GPU version will be campatible with both CPU and GPU.
2. tensorflow version change to >=1.2.0
The 'tf.clip_by_value' method only exists in tensorflow>=1.2.0


kimdwkimdw:
👍

jacknlliu:
@joschu Could you merge this for compatibility?

PR #52, opened 2017-07-11
Effectively apply weights from the replay buffer

farbeiza:
It seems that the weights retrieved from the replay buffer are not applied when training the model. Is there any reason for that or am I missing something?

In any case, I have added a parameter in order for them to be used; just in case it is useful.


siemanko:
Good catch!

PR #64, opened 2017-07-23
Use standardized advantages in trpo.

jhumplik:
The TRPO implementation calculates standardized advantages but then it doesn't use them. This fixes it.


Breakend:
I suggest this change be pulled in. The current TRPO results with the baseline code as is, are unsatisfactory. I think more work needs to be done to fix them, but this is at least one thing needed to align with the TRPO code.

joschu:
Wow, good catch. D'oh.

Issue #59, opened 2017-07-20
sum-tree unit test


miriaford:
According to the doc, `find_prefixsum_idx` method should return `the highest index i in the array such that sum(arr[0] ... arr[i-1]) <= prefixsum`.

If this is true, shouldn't the test return 4 instead of 3? 
https://github.com/openai/baselines/blob/master/baselines/common/tests/test_segment_tree.py#L44

and here, the test should also return 4 instead of 3
https://github.com/openai/baselines/blob/master/baselines/common/tests/test_segment_tree.py#L60

When `i == 4`, `sum(arr[0]+arr[1]+arr[2]+arr[3] <= 4.0` holds.

siemanko:
I think that's correct. I'd gladly accept a pull request with a fix that makes API and tests consistent.

miriaford:
@siemanko the edge case is a singular case, so I added a simple if statement on top. https://github.com/openai/baselines/pull/69

PR #69, opened 2017-07-25
SumSegmentTree edge case fix

miriaford:
As explained in https://github.com/openai/baselines/issues/59


siemanko:
Hmm... I think they way it is used here we also need to make sure that index is not out of range:
https://github.com/openai/baselines/blob/958810ed1e78624c300e327a0c79212f2453cfb7/baselines/deepq/replay_buffer.py#L112

PR #72, opened 2017-07-26
Fix pposgd and trpo_mpi import module issue

Pengchengpcx:
I have a problem that I cannot find modules trpo and ppo after pip install. I use virtualenv for python 3.6 in mac os10.11. I think add '__ init__' files into pposgd and trpo_mpi folders will add the two module into python package when do pip install. I notice that in deepq and other modules there exist the '__init__' files.


joschu:
Fixed

PR #75, opened 2017-07-27
Parameter space noise for DQN and DDPG

matthiasplappert:
- Adds a DDPG implementation with parameter space noise
- Adds parameter space for DQN (and related algorithms like DDQN, ...)


dmuestc:
is there a typo about dqn parameter space noise in file "baselines/deepq/build_graph.py" , line 226~227?  it seems that the input params of that func("perturb_vars(original_scope, perturbed_scope)") is useless...

matthiasplappert:
Yes, good catch. Would you like to open a PR to fix it?

PR #95, opened 2017-08-10
episode_rewards stored is by queue instead of list to improve memory …

cxxgtxy:
…efficiency when episode numer is large.


pzhokhov:
fixed upstream, closing this

Issue #74, opened 2017-07-26
Save and load of the trained TRPO and PPO agents


ViktorM:
Hi,

How save, load and visualization of the trained agents with TRPO or PPO algorithms can be done?

ashwinreddy:
Hello,

I think I've figured out how to do it, so I thought I would describe the process and possibly submit a pull request so other people can get started experimenting. 

The `run_atari.py` and `run_mujoco.py` files will go through the training/learning process and display some running statistics, but it does not save the network's model or parameters when the script completes. Luckily, you can use the `tf.train.Saver` class to do just that. It's as easy as instantiating the saver and then calling the save method with a session and file prefix.

Then, in another file, you'll want to instantiate the same policy (the default is a `MlpPolicy`) with the same parameters (hidden size, hidden layers, etc.). Then, use the saver class to restore the model, and voilá, you can call the act function and get back actions. 

This solution is a little janky and could probably be improved using the `ActWrapper` class that the DQN method uses.

joschu:
Thanks for the explanation, Viktor. I also use the tf.train.Saver, however, it's not in the released code because I figured that people would probably want to roll their own solutions.

dattatreya303:
Can we do this in pposgd_learn.py? I want to save the model after every 100 iterations, say. Would simply calling U.save_state(fname) work there?

zishanahmed08:
@ashwinreddy  Hi Ashwin,

I referred to your commit and am getting the below error suggesting that there is a mismatch in the size and dimensions somehow changes from 22528 to 3136
The only change from your code  I do is to replace mlp policy with cnn_policy since the run_atari trains the Pong env on Cnn_policy.
Any idea whats going wrong?

**InvalidArgumentError (see above for traceback): Assign requires shapes of both tensors to match. lhs shape= [22528,512] rhs shape= [3136,512]
	 [[Node: save/Assign_7 = Assign[T=DT_FLOAT, _class=["loc:@pi/lin/w"], use_locking=true, validate_shape=true, _device="/job:localhost/replica:0/task:0/device:CPU:0"](pi/lin/w, save/RestoreV2_7)]]**

BlazingFire:
I am also facing tensor shape mismatch issue, anyone got a workaround?

zishanahmed08:
@BlazingFire -The error for me was arising because i had missed one of the wrappers. It think it was the Frame Stack wrapper which stacks the 4 images. Hence the mismatch in dimensions.

[...]

PR #96, opened 2017-08-10
Save, Load, & Visualize PPO results

ashwinreddy:
As discussed in Issue #74, the code given for PPO trains the agent without saving its results, which can be confusing for newcomers. This pull requests adds two files that allow someone to train a PPO-based agent on the CartPole task and then load up the model and see the agent playing in real time. I also added a README file so that instructions on how to use the code can be clearly stated.

## Improvements
I'm not sure strictly how much of the setup from the training process needs to be duplicated in the visualizing process, but I included all of it so that it doesn't break.

It would also be more user-friendly if the training process returned an `ActWrapper`, but I wasn't sure how to do that.

### DIFF
--- a/baselines/pposgd/README.md
+++ b/baselines/pposgd/README.md
@@ -1,0 +1,8 @@
+## If you are curious.
+
+#### Train a Cartpole agent
+
+```bash
+python run_cartpole.py
+python enjoy_cartpole.py
+```
--- a/baselines/pposgd/enjoy_cartpole.py
+++ b/baselines/pposgd/enjoy_cartpole.py
@@ -18,17 +18,16 @@ def enjoy(env_id, num_timesteps, seed):
         return mlp_policy.MlpPolicy(name=name, ob_space=ob_space, ac_space=ac_space,
             hid_size=64, num_hid_layers=2)
     env = bench.Monitor(env, osp.join(logger.get_dir(), "monitor.json"))
+    obs = env.reset()
     env.seed(seed)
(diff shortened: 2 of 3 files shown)


hamzamerzic:
You introduce a dependency on roboschool, which is not needed in this case. 

Ideally we should be able to load any policy without needing to specify the entire graph structure. One approach for this would be to have a consistent naming for output tensor of each policy, and then simply import the graph metadata, as shown [here](https://stackoverflow.com/a/40765759/4330168). To load the policy you could then do something like:

```python
# Load the previous graph.
loader = tf.train.import_meta_graph(model_meta_path)

# Handle of the loaded graph.
g = tf.get_default_graph()

pi = g.get_tensor_by_name("policy_output:0")
with tf.Session() as sess:
    loader.restore(sess, tf.train.latest_checkpoint(model_dir_path))
    act = sess.run(pi, feed_dict={'observation:0': obs, 'stochastic:0': False, ...})  # Required tensor inputs...
```

longmoc:
How can I load and use the policy?

20chase:
@longmoc  https://github.com/20chase/baselines You can reference my implementation : )

NourBesbes:
@ashwinreddy When trying your solution it seems like I can't get into the loop : 
```
while not done: 
action = pi.act(True, obs)[0]
        obs, reward, done, info = env.step(action)
        env.render()
```

I can't see why ! Is anything wrong with the code ?

Issue #112, opened 2017-08-23
configure logger before use


andytwigg:
Call `logger.configure()` before use otherwise `logger.get_dir()` is None

joschu:
Fixed

PR #113, opened 2017-08-23
#112 configure logger before use

andytwigg:


joschu:
I've fixed this in the last commit. Thanks anyway for the PR though.

PR #119, opened 2017-08-28
fix _demo() in logger, patch for python2 support

kimdwkimdw:
Baselines don't support python2 officially, but small improvement makes better readability and python2 support.

I Love Python3, sometimes I need to write up python 2 code.

Opensim works only python2
ex) https://www.crowdai.org/challenges/nips-2017-learning-to-run

### DIFF
--- a/baselines/common/dataset.py
+++ b/baselines/common/dataset.py
@@ -47,7 +47,7 @@ class Dataset(object):
         return Dataset(data_map, deterministic)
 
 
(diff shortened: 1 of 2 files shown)


pzhokhov:
At this point, a) python2 is about to retire (https://pythonclock.org/) and b) there is a bunch more code in baselines that will require changes to work with python2, so I am going to close this.

PR #122, opened 2017-08-29
Added a script to play Breakout with an acktr checkpoint.

Bleyddyn:
Although this runs, my checkpoint file doesn't play very well at all. I think I may not be handling states and observations correctly.


emansim:
If your checkpointing doesn't work well, I replace save and load functions in acktr_disc.py code (same thing would work in acktr_cont.py) with the following code that also save other variables such as second order momentum etc. necessary for continuing training 

To save create tensorflow saver and use that saver to save model
```
saver = tf.train.Saver()

def save(saver, save_path):
    model_path = os.path.join(save_path, "model.ckpt")
    saver.save(sess, model_path)
```

To load create tensorflow saver and use it to restore the model
```
saver = tf.train.Saver()
saver.restore(model.sess, os.path.join(load_path, "model.ckpt"))
print ("Loaded Model")
```

PR #124, opened 2017-08-31
set num_cpu as config option in ACKTR run_atari

inoryy:


Eric-Wallace:
This change looks good to me.

PR #132, opened 2017-09-11
Bug fix in saving a2c model.

bhatiaabhinav:
Calling make_path(save_path) creates save_path as a folder instead of a file.
But joblib.dump expects save_path to be a file. Thus it was throwing error.
Changing make_path's argument to osp.dirname(save_path) fixes the problem.

Note: Tested on Windows 10. Python 3.5.4.


araffin:
This issue is also valid for ACER (#301), the same change should solve it too.

PR #153, opened 2017-10-02
Flush temporary file before compressing it.

akashin:
We need to flush the file buffer after `pickle.dump`, otherwise the resulting zip archive might be incomplete (reproducible, if the state consists of a single integer).


olegklimov:
merged

PR #154, opened 2017-10-03
added opencv as dependency to package

sbarratt:
baselines.common.atari_wrappers_deprecated.py has the line:

`import cv2`

Thus `opencv-python` is a dependency.


pascalwhoop:
I am getting a failure on install without this. +1
`pip3 install opencv-python `

solved it for me so I think the man is right!

PR #181, opened 2017-10-25
remove ref to simple_bench + rank variable not used [mpi gone]

JonathanRaiman:
Two minor fixes to baselines:
- remove reference to simple_bench
- remove reference to rank


abhigenie92:
Any workarounds? There is source now to include now either in sys path also.

PR #191, opened 2017-11-03
Update references to wrap_dqn

### LINKED ISSUE #190: ModuleNotFoundError: No module named 'baselines.common.atari_wrappers_deprecated'
This method is referenced in three places:

1. https://github.com/openai/baselines/blob/bb403781182c6e31d3bf5de16f42b0cb0d8421f7/baselines/deepq/experiments/atari/enjoy.py#L15
2. https://github.com/openai/baselines/blob/bb403781182c6e31d3bf5de16f42b0cb0d8421f7/baselines/deepq/experiments/atari/wang2015_eval.py#L10
3. https://github.com/openai/baselines/blob/bb403781182c6e31d3bf5de16f42b0cb0d8421f7/baselines/deepq/experiments/atari/train.py#L25

Should this instead be?
```
from baselines.common.atari_wrappers import wrap_deepmind
```

adammcnamara:
## Description

Updates references to removed package `from baselines.common.atari_wrappers_deprecated import wrap_dqn` to `from baselines.common.atari_wrappers import wrap_deepmind`

Closes #190

### DIFF
--- a/baselines/deepq/experiments/atari/enjoy.py
+++ b/baselines/deepq/experiments/atari/enjoy.py
@@ -12,7 +12,7 @@ from baselines.common.misc_util import (
     boolean_flag,
 )
 from baselines import bench
-from baselines.common.atari_wrappers_deprecated import wrap_dqn
+from baselines.common.atari_wrappers import wrap_deepmind
 from baselines.deepq.experiments.atari.model import model, dueling_model
 
 
@@ -31,7 +31,7 @@ def parse_args():
 def make_env(game_name):
     env = gym.make(game_name + "NoFrameskip-v4")
     env = bench.Monitor(env, None)
-    env = wrap_dqn(env)
+    env = wrap_deepmind(env)
     return env
 
 
--- a/baselines/deepq/experiments/atari/train.py
+++ b/baselines/deepq/experiments/atari/train.py
@@ -22,7 +22,7 @@ from baselines.common.misc_util import (
 )
 from baselines.common.schedules import LinearSchedule, PiecewiseSchedule
 from baselines import bench
-from baselines.common.atari_wrappers_deprecated import wrap_dqn
+from baselines.common.atari_wrappers import wrap_deepmind
 from baselines.common.azure_utils import Container
 from .model import model, dueling_model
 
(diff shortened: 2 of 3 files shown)


coderextreme:
Not sure what the intention is, but there's also baselines.deepq.wrap_atari_dqn(env) which calls wrap_deepmind with additional parameters. If no additional parameters are needed for the introduction of wrap_deepmind into the rest of baselines.deepq, it seems this patch may be okay.   If additional parameters are needed, then this patch needs updating to replace wrap_deepmind calls with something else (additional parameters?  change to deepq.wrap_atari_dqn?  Go back to the now non-existent deprecated?).  I just started looking at this, so I really have no clue. I am just interested in getting this thing fully patched.  Thanks for the patch, Adam, otherwise, I might have been tempted to submit a patch myself.  I suspect you're quite a bit more familiar with this than I am.

kdu4108:
Hm, I implemented these changes and tried running the given visualization script for a pretrained Atari agent, `python -m baselines.deepq.experiments.atari.enjoy --model-dir /tmp/models/model-atari-duel-pong-1 --env Pong --dueling`. 
It threw this error:
`InvalidArgumentError (see above for traceback): Assign requires shapes of both tensors to match. lhs shape= [8,8,1,32] rhs shape= [8,8,4,32]`
I'm pretty new to deep learning, and while I don't think these errors are caused by these changes, any help would be appreciated!

coderextreme:
Sorry, I don't have a clue.  Perhaps the author of the patch can help.

kdu4108:
John, thanks for a prompt response--did you run into a similar problem?

coderextreme:
No, after my code didn't run I stopped because I was unsure of correct
patch.

coderextreme:
Have you tried calling wrap_atari_dqn instead?

kdu4108:
What happened when you tried to run your code? And no, since `from baselines.common.atari_wrappers_deprecated import wrap_dqn` and `from baselines.common.atari_wrappers import wrap_deepmind` seem analogous whereas
wrap_atari_dqn comes from deepq/__init__.py, which seems like something else.

coderextreme:
If you look closely, wrap_atari_dqn calls wrap_deepmind with extra parameters.

John

Sent from Mail for Windows 10

pzhokhov:
I think this is obsolete, closing it.

PR #193, opened 2017-11-05
Add OpenCV to install dependencies

ryanjulian:
OpenCV is necessary to run atari demos


pzhokhov:
added already (with open version), closing this

PR #199, opened 2017-11-12
Put train_pong back after being deleted

smc77:
I replaced train_pong.py, but updated it for the renamed function deepq to deepmind.


pzhokhov:
obsolete, closing

PR #214, opened 2017-11-18
Make reward buffer length consistent across PO algorithms.

hamzamerzic:
Reward buffer length for [PPO](https://github.com/openai/baselines/blob/master/baselines/ppo1/pposgd_simple.py#L139) and [PPO2](https://github.com/openai/baselines/blob/master/baselines/ppo2/ppo2.py#L180) is 100, while for TRPO it is [currently 40](https://github.com/openai/baselines/blob/master/baselines/trpo_mpi/trpo_mpi.py#L183). 

This inconsistency means that the EpRewMean is smoother in case of PPO algorithms compared to TRPO, which can be slightly misleading if we compare learning in terms EpRewMean, or track EpRewMean during training.


pzhokhov:
closing this, but a PR with the same idea to reduce conflicts: https://github.com/openai/baselines/pull/712

PR #216, opened 2017-11-20
Fix ppo1 reward statistic.Usually, unclipped reward is used in publis…

cxxgtxy:
…hed paper.

Adjust hyper parameter for ppo1 based on the original paper.


cxxgtxy:
Hi, anyone check it, please.

unixpickle:
The unclipped rewards are stored in the monitor files created by the `Monitor` wrapper. See https://github.com/openai/baselines/issues/232.

PR #231, opened 2017-12-13
make trpo returns consistent with ppo returns

bjkomer:
Currently the calculation for returns is different between these two algorithms due to different buffer lengths. The same experience given to both algorithms logs different results. This change will allow comparisons to be more valid.


unixpickle:
The unadulterated rewards are stored in the monitor files created by the Monitor wrapper. See #232.

PR #233, opened 2017-12-16
A2C, PPO2, ACKTR refactor (WIP)

kengz:


Sohojoe:
I am getting some excellent results with acktr so glad it is getting some love! 

@kengz - I'm not sure if you are looking for feedback but these are the areas that caused me the most pain and where I had to refactor code to cover my use cases. 

1) It would be great to have support for all 4 of these modes out of the box:
        * PixelObsToDiscreteAction
        * ContinuousObsToDiscreteAction
        * PixelObsToContinuousAction
        * ContinuousObsToContinuousAction

Currently I'm using ContinuousObsToDiscreteAction & ContinuousObsToContinuousAction. It was painful to extend, and my code is a mess. I will be using PixelObsToDiscreteAction & PixelObsToContinuousAction at some point too

2) Make nstack, num_hid_layers, hid_size  standard options for all modes. 

For me, been able to grid search these params was a gamechanger for optimisation. 

Currently, num_hid_layers and hid_size are hard coded at 64 x 2. nstack is implemented in atari but is hardcoded as 2 in mujoco. 

3) Standardise Save / Load for all modes (is not implemented in mujoco). Ideally, implement the checkpoint_freq from DQN (where it saves the best model)

ViktorM:
Hi @kengz! It's a very important work. I have the same suggestions as commentator above - to support a continuous control for ACKTR in the same way as discrete - using Model class. And to add options to customize layer sizes and activation functions, as well as save/load out of the box.

kengz:
Hi, thanks for the suggestion. Good to see the potential problems this will solve. I'll try to get back to this soon when time permits.

PR #265, opened 2018-01-25
fix logger error for trpo_mpi

20chase:
Can't run logger.configure() if rank != 0


20chase:
Hi @joschu 

trpo_mpi cannot run with mpi4py because logger.configure() assert rank=0.
so I modify the run_mujoco.py according to ppo1 to make it work. : )

joschu:
why change the section with argparse?

20chase:
sorry, I have fixed it.

joschu:
Cool, merged

PR #268, opened 2018-01-27
Fix fc call in AcerLstmPolicy

ei-grad:
The `act` keyword was removed from baselines.a2c.utils.fc in commit 9fa8e1b.


ei-grad:
@joschu

PR #281, opened 2018-02-07
Fixed name of default MuJoCo environment

atamahjoubfar:
Gym environment "Reacher-v1" is retired. So, if a MuJoCo environment is not specified in the arguments, and the code is run for the default environment, it would not work. To resolve the issue the default environment is changed to "Reacher-v2".


pzhokhov:
should be fixed now, closing this

PR #306, opened 2018-03-01
Fix bug in DDPG parameter space noise adaptation

daniel-ziegler:
The variable `t` here should be `t_train`. `t` is invariant for this loop, so `t % param_noise_adaption_interval == 0` will evaluate to the same thing each time. For the default values of the parameters `param_noise_adaption_interval` (50) and `nb_rollout_steps` (100), this causes the code to adapt the scale of the parameter space noise in every single training step, which is correct but slow.

(For other values of the parameters, the adaptation won't happen for long periods of time, which is a problem.)


matthiasplappert:
Good catch!

Issue #301, opened 2018-02-27
Error Saving ACER/A2C Model


araffin:
Hello,

EDIT: this bug is also valid for a2c

There is a bug in the `save()` method of ACER.
https://github.com/openai/baselines/blob/master/baselines/acer/acer_simple.py#L201
```python
def save(save_path):
        ps = sess.run(params)
        make_path(save_path)
        joblib.dump(ps, save_path)
```

`make_path()` is creating a folder whereas `joblib.dump()` is expecting a path to a file.
That results in a `IsADirectoryError`.

One simple fix would be to remove the `make_path()` line. You should also document what is expected as `save_path` parameter (is that a path to a folder, a pickle file, an archive ?).

araffin:
#132 Pull request for A2C solving this issue

araffin:
@unixpickle Should I submit a pull request to solve the same issue (#132) for ACER ?

unixpickle:
@araffin sure. Ideally we'll get rid of the duplicated code at some point, but for the time being we might as well fix it in all places.

araffin:
Pull request #308 submitted ;)

PR #312, opened 2018-03-02
Add missing import in ACER

araffin:
Fixes the missing import bug introduced in #308.


araffin:
@unixpickle as discussed in #301, the fix for ACER was the same as for A2C, however, there was a missing import in the ACER code.

araffin:
Solved by #385

Issue #364, opened 2018-04-13
PPO2 - tuple index out of range when using Discrete action state. 


redknightlois:
I have a gym.spaces.Discrete(3) action space on my environment but when I try to execute PPO2 latest code (built from source), I get:
```
Traceback (most recent call last):
  File "E:/Src/trendstone-git/src/learner-v2/train.py", line 143, in <module>
    total_timesteps=max_timesteps)
  File "C:\Anaconda3\lib\site-packages\baselines\ppo2\ppo2.py", line 178, in learn
    model = make_model()
  File "C:\Anaconda3\lib\site-packages\baselines\ppo2\ppo2.py", line 173, in <lambda>
    max_grad_norm=max_grad_norm)
  File "C:\Anaconda3\lib\site-packages\baselines\ppo2\ppo2.py", line 16, in __init__
    act_model = policy(sess, ob_space, ac_space, nbatch_act, 1, reuse=False)
  File "C:\Anaconda3\lib\site-packages\baselines\ppo2\policies.py", line 135, in __init__
    actdim = ac_space.shape[0]
IndexError: tuple index out of range
```

Given that `gym.spaces.Discrete` does not implement `.shape` I workaround it creating an inherited class and returning `(1,)` but I am wondering if I am not making things wrong. 

EDIT: Moreover now it will only sample actions as 0 and 1. When I put (2,) it will sample between 0 and 3.  Any idea why this is not happening would be welcome.

unixpickle:
It looks like MlpPolicy is designed for continuous action spaces, and CnnPolicy is designed for discrete action spaces. It should be fairly easy to take the logic from CnnPolicy and move it to MlpPolicy. E.g. change `actdim = ac_space.shape[0]` to `actdim = ac_space.n`.

pzhokhov:
this should be addressed now, I'll close the issue. @redknightlois please reopen if you still experience the problem.

Issue #357, opened 2018-04-11
Initial observations are zero in A2C 


filipemtz:
In baselines/a2c/a2c.py, lines 98-100 are:
        self.obs = np.zeros((nenv, nh, nw, nc), dtype=np.uint8)
        self.nc = nc        
        obs = env.reset()

Shouldn't them be:
        self.obs = np.zeros((nenv, nh, nw, nc), dtype=np.uint8)
        self.nc = nc        
        self.obs[:] = env.reset()

as in PPO? 

In the current version, the obs variable is local and it's not used anywhere else in the code. 
The effect of this is that the first observation of the first episode will always be zero.

By the way, why are there copies of the same classes across the implementations of different methods? For example, the code above is part of the Runner class which is present in both, A2C and PPO. Is it to simplify the extraction of the code of a method from the rest of the code?

unixpickle:
This does appear to be a bug, which is fixed in PPO.

The code duplication is not ideal (as you can see, since it leads to bug like this).

pzhokhov:
the code duplication should be addressed now, closing the issue

Issue #345, opened 2018-03-31
PrioritizedReplayBuffer - assert alpha > 0?


RylanSchaeffer:
I'm confused why the `PrioritizedReplayBuffer` has the following statement:

`assert alpha > 0`

Why is this not `assert alpha >= 0`? I thought `alpha=0` is a permissible value that corresponds to uniform random sampling.

pzhokhov:
@RylanSchaeffer you are right. We have changed the assert to alpha >= 0 and added some tests, closing the issue.

PR #385, opened 2018-05-01
import rl-algs

pzhokhov:
should address https://github.com/openai/baselines/issues/364, https://github.com/openai/baselines/issues/357, https://github.com/openai/baselines/issues/345,

### DIFF
--- a/baselines/acer/acer_simple.py
+++ b/baselines/acer/acer_simple.py
@@ -5,6 +5,7 @@ import tensorflow as tf
 from baselines import logger
 
 from baselines.common import set_global_seeds
+from baselines.common.runners import AbstractEnvRunner
 
 from baselines.a2c.utils import batch_to_seq, seq_to_batch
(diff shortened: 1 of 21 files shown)


baselines/acer/acer_simple.py
```diff
@@ -13,6 +14,8 @@
 from baselines.a2c.utils import get_by_index, check_shape, avg_norm, gradient_add, q_explained_variance
 from baselines.acer.buffer import Buffer
 
+import os.path as osp
```
christopherhesse: that's funny, looks like this fix https://github.com/openai/baselines/commit/569bd426297859c57c0c2ba6482f1349ac12fc01 missed the import?

pzhokhov: I think so... did not seem to work without it

araffin: Closes https://github.com/openai/baselines/pull/312

pzhokhov: sounds good, but we'll have to export this to baselines.

PR #388, opened 2018-05-02
setup travis ci build

pzhokhov:


.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+
+services:
+    - docker
+
+install:
+    - pip install flake8
+    - docker build . -t baselines-test
+
+script:
+    - flake8 --select=F common
```
christopherhesse: How does this run against baselines/common instead of common?

christopherhesse: How does this run against baselines/common instead of common?

pzhokhov: good point, it should be baselines/common. Fixed now, it still should run clean.

PR #401, opened 2018-05-09
Fix a bug in data loading

ywchao:
Currently the for loop only processes one entry in the array, causing the flattened array to have all zeros after the first entry.


pzhokhov:
fixed upstream using np.reshape directly

PR #418, opened 2018-06-03
Fix DummyVecEnv reusing buffer for observation

AdamGleave:
DummyVecEnv historically reused buffers for the values returned by step(). This differs from the semantics of SubprocVecEnv and can lead to surprising results. @unixpickle fixed this in 8b3a6c20519a45a1387c7d23b98b55d5961f3403, but part of the change appears to have been clobbered in 69f25c6028d71e23a7b267b0e1071ad216290f91 such that the current behavior is that the _observation_ buffer is reused while reward, done and info are copied. This PR fixes this to copy the observation buffer as well. Requesting review from @unixpickle or @pzhokhov.

### DIFF
--- a/baselines/common/vec_env/dummy_vec_env.py
+++ b/baselines/common/vec_env/dummy_vec_env.py
@@ -62,6 +62,6 @@ class DummyVecEnv(VecEnv):
 
     def _obs_from_buf(self):
         if self.keys==[None]:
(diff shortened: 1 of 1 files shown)


pzhokhov:
should be fixed now by virtue of using copy_obs_dict method; closing this.

Issue #414, opened 2018-05-29
ppo2 (env humanoid-v2) does not work with default parameters


foolyc:
with default parameters
![ppo2](https://user-images.githubusercontent.com/7566074/40637837-2897afcc-633a-11e8-97d5-c02aae6905d2.jpg)

And [ppo2.py](https://github.com/openai/baselines/blob/master/baselines/ppo2/ppo2.py) Runner.run() will run self.nsteps steps in every episode without early terminal, the sample efficiency will be very low.

pzhokhov:
Humanoid can be a bit tricky to train. We are publishing training hyperparameters for ppo (although for ppo1, not ppo2) for humanoid in this PR: https://github.com/openai/baselines/pull/422/files, file ppo1/run_humanoid.py. PR should go in soon. You can try the same hyperparameters with ppo2, in principle, it should work. Eventually, I'll get around to try that too, and will publish the results

foolyc:
@pzhokhov  Thanks for your reply. 
Additionally
1) Could you add a ppo demo with the LSTM policy? The origin code in ppo2 with LSTM does not work well.
2) Could you add early terminal in the ppo2 sampling to increase the sample efficiency?

jiameij:
@foolyc hi~
Do you have tried baselines/ppo1/run_humanoid ? how many total_timesteps needed for training humanoid-v2 so that it can run?
thank you!

foolyc:
@jiameij Hi
I have not test the ppo1 yet, but usually it needs at least n * 1e4 episodes to learn the rl task.

jiameij:
hi~
i test ppo1.run_humanoid, the result of results_plotter.py as follows:
![2e8](https://user-images.githubusercontent.com/35295912/41638952-f2d2fc7e-7429-11e8-8de2-c01330335500.png)
then i sample trajectory and render it via 
`
expert_path = "/home/jmj/Downloads/stochastic.Humanoid-v2.pkl"
with open(expert_path, "rb") as f:
    traj_data = pkl.load(f)
obs = []
acs = []
rets = []
lens = []
for traj in traj_data:
    acs = traj["ac"]
    for ac in acs:
        env.step(ac)
        env.render()
        time.sleep(0.02)
`  
it seems that the Humanoid-v2 model doesn't learn to run. it runs about 4 steps and fall down.
should i train the model more timesteps or is there other tricks?
thank you !

PR #422, opened 2018-06-06
Import internal changes

pzhokhov:
Release internal changes. Addresses this issue: https://github.com/openai/baselines/issues/414


andytwigg:
It looks like this commit is missing `baselines.common.tile_images`

For example, importing SubprovVecEnv fails:
```
from baselines.common.vec_env.subproc_vec_env import SubprocVecEnv
>> from baselines.common.tile_images import tile_images
>> ModuleNotFoundError: No module named 'baselines.common.tile_images
```

pzhokhov:
yes, thanks for pointing that out! Will be fixed in the PR: https://github.com/openai/baselines/pull/427

PR #437, opened 2018-06-11
Fix dtype for wrapper observation spaces

kcobbe:


unixpickle:
Actually, I realized that doing `self.observation_space.dtype = np.float32` isn't good, because you're just mutating the observation space of the input environment. You should probably construct a new box space, e.g. `gym.spaces.Box(low=0, high=1, shape=env.observation_space.shape, dtype=np.float32)`

skpenn:
It's a terrible bug which will cause DQN fails in Atari games. Actually In baselines/common/input.py line 18:
`input_x = tf.placeholder(shape=input_shape, dtype=ob_space.dtype, name=name)`
Leads to the data type of the observation's tf placeholder mismatch with the real data's. And model fails to train.
So why nobody merges this pull request into main branch

pzhokhov:
effectively megred (I merged it in together with the refactor), closing this

PR #444, opened 2018-06-16
Remove README section on download pretrained model

pgericson:
The described code was removed in edb52c2


pgericson:
the download section was removed in a different commit. will close this one

PR #447, opened 2018-06-20
GAIL: bugfix in dataset loading

AdamGleave:
Currently all but the first episode are zeroed out due to a whitespace error.

I'm confused why ```flatten``` has been implemented with a for-loop at all, I think it can be replaced directly with a NumPy reshape command. But I don't understand enough of the surrounding code to be confident this wouldn't break anything.

Request review from @andrewliao11

### DIFF
--- a/baselines/gail/dataset/mujoco_dset.py
+++ b/baselines/gail/dataset/mujoco_dset.py
@@ -56,7 +56,7 @@ class Mujoco_Dset(object):
             for l, x_i in zip(episode_length, x):
                 y[start_idx:(start_idx+l)] = x_i
                 start_idx += l
-                return y
+            return y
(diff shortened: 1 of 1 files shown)


andrewliao11:
Just tested it. Yes, that's a bug and it should be implemented as something like:
```python
self.obs = np.reshape(obs, [-1, np.prod(obs.shape[2:])])
self.acs = np.reshape(acs, [-1, np.prod(acs.shape[2:])])
```
could you modify it and send the push again?

AdamGleave:
Sure, done.

andrewliao11:
@pzhokhov @joschu could you merge this? should be independent to other algos

pzhokhov:
looks good to me

PR #461, opened 2018-07-09
Update subproc_vec_env.py

oeclint:
Allow worker to be defined externally. If SubprocVecEnv is subclassed new attributes in the subclass can now be accessed if defined in the worker.


pzhokhov:
@oeclint could you provide an example of that? I am not sure I understand the use case for it.

pzhokhov:
closing due to age

PR #488, opened 2018-07-29
Ability to choose between tensorflow and tensorflow-gpu

fabito:
This PR adds the ability to choose between tensorflow gpu or non gpu flavors as suggested here: https://github.com/tensorflow/tensorflow/issues/7166#issuecomment-280881808.
It basically moves the tensorflow dependency from `install_requires` to `extra_requires`.

Examples:

```bash
# no tensorflow
pip install git+https://github.com/openai/baselines.git#egg=baselines

# tensorflow
pip install git+https://github.com/openai/baselines.git#egg=baselines[tf]

# tensorflow-gpu
pip install git+https://github.com/openai/baselines.git#egg=baselines[tf_gpu]
```


pzhokhov:
since there is a desire to use a variety of builds of tensorflow (non-gpu and gpu, nightly, custom builds); we decided to remove tensorflow from explicit dependencies entirely, and fail at the install time if no version of tensorflow is present or if it is below 1.4. Burden of installing preferred tensorflow falls on user :)

PR #490, opened 2018-07-31
refactor a2c, acer, acktr, ppo2, deepq, and trpo_mpi

pzhokhov:
Refactor of a2c, acer, acktr, ppo2, deepq, and trpo_mpi that provides:
1. common network building and policy API
2. common serialization API 
3. common runner for atari, mujoco and retro (remove  multiple run_{atari, mujoco} files that have different defaults) that allows for MPI and subprocess parallelization 
4. unified tests
5. docs (docstrings, readmes)


olegklimov:
Yes, I like this patch and it is a monstrous effort to do de-duplication. This patch is pretty big and yet Runner still seems to be not unified enough (I guess because you'd have to rename all the variables in algorithms). But definitely a move in right direction.

Random thoughts:

AppendTimeout will not work with any algorithm in baselines, because it produces dict observations.

A way to get rid of a2c/utils.py is to use Tensorflow equivalents, such as tf.layers.conv2d(). If we finally have tests to catch regressions, this seems doable.

CubeCrash-v0 and MemorizeDigits-v0 are designed for unit tests.

baselines/run.py seems to be very important for this kind of refactoring, but also seems to be missing a lot of things now, but again a step in right direction.

Can this be merged in its current stage -- I don't know, likely yes it can from my point of view, if you have unit tests running.

pzhokhov:
Thanks @olegklimov ! 
Did not know about CubeCrash-v0 and MemorizeDigits-v0, I'll look into those. 
replacing a2c.utils with tf equivalents - sounds like a good idea to me, but, ideally, test coverage needs to be more complete (i.e. right now there are unified tests for a2c, acer, acktr_discrete, deepq, ppo2 and trpo_mpi, but virtually no tests for other algorithms); so I'd prioritize that after converting other algorithms to the same API. 
AppendTimeout - noted; do you think we should make move dict observation spaces into the baselines as well? Do you have some simple example that illustrates usage of AppendTimeout or other dict observation space?

olegklimov:
> dict observation spaces into the baselines

It is questionable whether or not it should be in baselines -- makes code more complicated. I think the idea of baselines is to be as simple as possible.

OTOH, all new environments we're likely to create will use dict. (released robotics envs already do)

baselines/common/running_mean_std.py
```diff
@@ -13,20 +16,71 @@ def update(self, x):
...
+    
+    return new_mean, new_var, new_count
+    
+
+class TfRunningMeanStd(object):
+    # https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
+    '''
+    TensorFlow variables-based implmentation of computing running mean and std
+    Benefit of this implementation is that it can be saved / loaded together with the tensorflow model
+    '''
```
pzhokhov: this implementation, although allows saving and restoring, works much slower than the RunningMeanStd; and when used in VecNormalize with mujoco, drops frame rate by a factor of 3 :/ Any ideas on how can we speed it up?

olegklimov: In principle, you only need to scale observations (and reward) once you've finished sampling. This way it won't require any speed. But the downside is then all algorithms have to be modified, such as scale already saved observations in runner.

wookayin:
Awesome! Do you have any benchmark result publicly available (or, comparison to the previous)?

pzhokhov:
Yeah, good idea! We may be able to publish the benchmark results.

pzhokhov:
https://github.com/openai/baselines/pull/502

PR #513, opened 2018-08-16
Adding SubprocVecEnv remote interface for grabbing frames

dineshj1:
SubprocVecEnv currently offers a 'render' remote command that allows on-screen rendering. This PR implements a 'grab_frame' remote command to grab a frame without rendering on-screen.


pzhokhov:
not sure I understand why this is needed - is this use case not covered by render with `mode='rgb_array'`?

PR #527, opened 2018-08-20
prioritized experience replay bug

XFFXFF:
The implementation of prioritized experience replay is a bit wrong. To smaple a minibatch of size k, the range [0, p_total] is divided equally into k ranges. Next, a value is uniformly sampled from each range. Finally the transitions that correspond to each of these sampled values are retrieved from the tree. 
Refer to the paper [Prioritized Experience Replay](https://arxiv.org/abs/1511.05952). See Appendix B.2.1 for more details


pzhokhov:
Looks legit to me... @siemanko, what do you think?

Issue #521, opened 2018-08-18
Error using baselines.results_plotter


DamienLancry:
Dear OpenAI and Community,

I am trying to figure out how to use the results_plotter function.
I have run a simple test with the atari environment Pong by running
```
python -m baselines.run --alg=a2c --env=PongNoFrameskip-v4 --num-timesteps=2e7 --save_path=~/models/pong_20M_ppo2
```
and then I ran
```
python -m baselines.results_plotter --dirs=/tmp/openai-2018-08-16-18-14-01-752470
```
And it works perfectly.
But then I tried:
```
python -m baselines.run --alg=ppo2 --env=HalfCheetah-v2 --num_env=8 --save_path=~/models/HC_1M_ppo2
```
then:
```
python -m baselines.results_plotter --dirs=/tmp/openai-2018-08-18-17-32-10-748103
```
And I get the following error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/homes/drl17/Documents/Project/baselines/results_plotter.py", line 87, in <module>
    main()
  File "/homes/drl17/Documents/Project/baselines/results_plotter.py", line 83, in main
    plot_results(args.dirs, args.num_timesteps, args.xaxis, args.task_name)
  File "/homes/drl17/Documents/Project/baselines/results_plotter.py", line 61, in plot_results
    ts = load_results(dir)
  File "/homes/drl17/Documents/Project/baselines/bench/monitor.py", line 119, in load_results
    df = pandas.read_csv(fh, index_col=None)
  File "/homes/drl17/Documents/Project/Torcs_Project/env3/lib/python3.5/site-packages/pandas/io/parsers.py", line 678, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "/homes/drl17/Documents/Project/Torcs_Project/env3/lib/python3.5/site-packages/pandas/io/parsers.py", line 446, in _read
    data = parser.read(nrows)
  File "/homes/drl17/Documents/Project/Torcs_Project/env3/lib/python3.5/site-packages/pandas/io/parsers.py", line 1036, in read
    ret = self._engine.read(nrows)
  File "/homes/drl17/Documents/Project/Torcs_Project/env3/lib/python3.5/site-packages/pandas/io/parsers.py", line 1848, in read
[...]

Issue #519, opened 2018-08-17
VecEnv SubProcEnv


DamienLancry:
Just wondering what is a vectorized environment? Where to get additional information about it?
Cheers!

pzhokhov:
Vectorized environment is a wrapper for multiple environments running in parallel (so that an algorithm can get updates from multiple trajectories at once - this gives larger update batches and reduces the variance of the updates). SubprocVecEnv uses environments running is separate subprocesses that communicate with head process via pipes. ShmemVecEnv is largely a duplicate of SubprocVecEnv, we'll be cleaning that up at some point. 
For theoretical / literature justification of sampling several trajectories at once, let's ask @joschu, @olegklimov or @unixpickle.

yoojinoh:
Hi @pzhokhov,  Thanks for your answer. But I don't understand how SubProcEnv different from DummyVecEnv.
I am trying to use PPO2 with a customized environment and what I noticed in the run.py the mujoco uses the SubProcVec when there are a number of environment copies being run in parallel or else just a DummyVecEnv.
Do I have to use them together or should the SubProcEnv do the job that the DummyVecEnv does?

pzhokhov:
Hi @yoojinoh! DummyVecEnv and SubprocVecEnv are doing the same job (in fact, in terms of functionality, one can be replaced with the other - they *should* have the same API defined by abstract superclass VecEnv), the only difference is that SubprocVecEnv runs steps in multiple environments in subprocesses, i.e. actually in parallel, whereas DummyVecEnv runs steps for multiple environments one after another and is not using any kind of parallelism. 
To elaborate on that point, consider a call to venv.step(), where venv is in instance of either DummyVecEnv or SubprocVecEnv. DummyVecEnv interates over multiple environments in a for loop, calling env.step() for the wrapped environments one after another. If using Nenv environments, and each of them takes time Tstep to make a step, DummyVecEnv.step will take Nenv * Tstep.   SubprocVecEnv first sends the data to all of the subprocesses, and then waits for all of them to send the data back - this way, computation of environment steps happens in parallel. This way, SubprocVecEnv only takes time max(Tstep_i) to run a step.  Natually, when using only one environment, SubprocVecEnv does not give any gain in speed (in fact, it introduces some inter-process communication overhead), that's why one num_envs == 1, we use DummyVecEnv.

simoninithomas:
[...]

PR #533, opened 2018-08-22
Fix result_plotters in vectorized mujoco environments

DamienLancry:
I was trying to plot the results of training of ppo2 on vectorized mujoco environments (Reacher) and got an error that I talked about in Issues #521 .
I managed to solve this issue that was caused by the fact that only one monitor.csv file is created even when --num_env= 8 (for example) is specified although there should be 8 files created as it is the case when running a training on an atari environment. 
This pull request would also solve another issue , which is that the seed is the same in all parallel mujoco environments (when a --num_env >1 is specified) because of the strange behavior of lambda functions in python  (see #529 ) . I could not do otherwise than to solve this other issue (seed) for fixing this one (csv). 
I made sure that the function  make_mujoco_env is not broken in acktr.run_mujoco . (the function is also used in ppo1 but i did not bother checking if it was broken or not since it is deprecated anyway. But I think it s still working anyway.)

### DIFF
--- a/baselines/common/cmd_util.py
+++ b/baselines/common/cmd_util.py
@@ -15,40 +15,28 @@ from baselines.bench import Monitor
 from baselines.common import set_global_seeds
 from baselines.common.atari_wrappers import make_atari, wrap_deepmind
 from baselines.common.vec_env.subproc_vec_env import SubprocVecEnv
+from baselines.common.vec_env.dummy_vec_env import DummyVecEnv
+from baselines.common.retro_wrappers import RewardScaler
 
-def make_atari_env(env_id, num_env, seed, wrapper_kwargs=None, start_index=0):
+
+def make_vec_env(env_id, env_type, num_env, seed, wrapper_kwargs=None, start_index=0, reward_scale=1.0):
     """
-    Create a wrapped, monitored SubprocVecEnv for Atari.
+    Create a wrapped, monitored SubprocVecEnv for Atari and MuJoCo.
     """
     if wrapper_kwargs is None: wrapper_kwargs = {}
     mpi_rank = MPI.COMM_WORLD.Get_rank() if MPI else 0
     def make_env(rank): # pylint: disable=C0111
         def _thunk():
-            env = make_atari(env_id)
+            env = make_atari(env_id) if env_type == 'atari' else gym.make(env_id)
             env.seed(seed + 10000*mpi_rank + rank if seed is not None else None)
             env = Monitor(env, logger.get_dir() and os.path.join(logger.get_dir(), str(mpi_rank) + '.' + str(rank)))
-            return wrap_deepmind(env, **wrapper_kwargs)
+            if env_type == 'atari': return wrap_deepmind(env, **wrapper_kwargs) if env_type == 'atari' else env
+            elif reward_scale != 1: return RewardScaler(env, reward_scale)
+            else: return env
         return _thunk
     set_global_seeds(seed)
-    return SubprocVecEnv([make_env(i + start_index) for i in range(num_env)])
-
-def make_mujoco_env(env_id, seed, reward_scale=1.0):
-    """
-    Create a wrapped, monitored gym.Env for MuJoCo.
-    """
-    rank = MPI.COMM_WORLD.Get_rank()
-    myseed = seed  + 1000 * rank if seed is not None else None
-    set_global_seeds(myseed)
-    env = gym.make(env_id)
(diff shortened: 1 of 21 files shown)


baselines/run.py
```diff
@@ -91,7 +91,7 @@ def build_env(args):
                                    inter_op_parallelism_threads=1))
 
         if args.num_env:
-            env = SubprocVecEnv([lambda: make_mujoco_env(env_id, seed + i if seed is not None else None, args.reward_scale) for i in range(args.num_env)])    
+            env = SubprocVecEnv([(lambda y: (lambda: make_mujoco_env(env_id, seed, y, args.reward_scale)))(i) for i in range(args.num_env)])
```
pzhokhov: why double lambda and not:
`[lambda: make_mujoco_env(env_id, seed, i, args.reward_scale) for i in range(args.num_env)] ` ?

baselines/common/cmd_util.py
```diff
@@ -32,15 +32,15 @@ def _thunk():
...
     """
     Create a wrapped, monitored gym.Env for MuJoCo.
     """
     rank = MPI.COMM_WORLD.Get_rank()
-    myseed = seed  + 1000 * rank if seed is not None else None
+    myseed = seed + i + 1000 * rank if seed is not None else None
     set_global_seeds(myseed)
     env = gym.make(env_id)
-    logger_path = None if logger.get_dir() is None else os.path.join(logger.get_dir(), str(rank))
+    logger_path = None if logger.get_dir() is None else os.path.join(logger.get_dir(), str(rank)+ '.' + str(i))
```
pzhokhov: it may make sense to make this more like make_atari_env (lines 19-33 in this file) and have num_env passed as an argument... In fact, most of the code from make_atari_env can be reused; so you can make a common make_vec_env function that does the magic with monitor, seeds etc, and accepts either wrap_deepmind (for atari) or gym.make(...) for mujoco. Does that make sense?

baselines/run.py
pzhokhov: ah, got it, because of this: https://stackoverflow.com/questions/6076270/python-lambda-function-in-list-comprehensions. ok. Still though, see the comment to cmd_util, maybe this can be addressed differently.

baselines/common/cmd_util.py
DamienLancry: I did it:
I unified make_atari_env and make_mujoco_env in one function make_vec_env. 
The easy way to do it was to pass env_type as an argument in make_vec_env.
The default behaviour for mujoco env is still to set num_env to 1 for mujoco and to 8 for atari. 
I noticed that when using nenv = 1, for atari, it is using SubProcVecEnv instead of DummyVecEnv which adds unnecessary process communication overhead as pointed out in #519 . So i corrected it. 
I am currently replacing make_mujoco_env by make_vec_env in acktr.run_mujoco .
I am also replacing make_mujoco_env by make_vec_env in ppo1.run_mujoco and ppo1.run_humanoid even though ppo1 is deprecated.
My ide automatically removed some unecessary spaces, i did not want to commit that, sorry about that. See the second commit instead of the third one.

DamienLancry:
See the second commit, I messed up with the following ones but in the end they don t change anything.
Sorry about the inconvenience.

pzhokhov:
looks good to me (except for the comment in cmd_util), although I'd prefer more traditional, indented if-elif-else instead of one-liners ^_^

baselines/common/cmd_util.py
```diff
@@ -15,22 +15,28 @@
...
     if wrapper_kwargs is None: wrapper_kwargs = {}
     mpi_rank = MPI.COMM_WORLD.Get_rank() if MPI else 0
     def make_env(rank): # pylint: disable=C0111
         def _thunk():
-            env = make_atari(env_id)
+            env = make_atari(env_id) if env_type == 'atari' else gym.make(env_id)
             env.seed(seed + 10000*mpi_rank + rank if seed is not None else None)
             env = Monitor(env, logger.get_dir() and os.path.join(logger.get_dir(), str(mpi_rank) + '.' + str(rank)))
-            return wrap_deepmind(env, **wrapper_kwargs)
+            if env_type == 'atari': return wrap_deepmind(env, **wrapper_kwargs) if env_type == 'atari' else env
```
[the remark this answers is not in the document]
DamienLancry: Oh yes sorry about it I forgot to erase it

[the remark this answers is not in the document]
DamienLancry: I wanted to erase everything after the return wrap_deepmind actually

DamienLancry:
There it is: I erased the useless if ... else ... !

PR #539, opened 2018-08-24
Fixes issue with pickling policies when SubprocVecEnv is used

gregorgebhardt:


baselines/ppo2/ppo2.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, *, policy, ob_space, ac_space, nbatch_act, nbatch_train,
         with tf.variable_scope('ppo2_model', reuse=tf.AUTO_REUSE):
             act_model = policy(nbatch_act, 1, sess)
             train_model = policy(nbatch_train, nsteps, sess)
+            eval_model = policy(1, 1, sess)
```
pzhokhov: why is eval_model needed?

pzhokhov:
from what I understand the PR intends to solve the issue with vecenvs not being serializable (whereas observation space and action space are). Could you provide a unit test that fails without this change and is fixed with it? Thanks!

baselines/ppo2/ppo2.py
gregorgebhardt: We have added that to evaluate the learned policy on a single instance of our environment

gregorgebhardt:
I reverted the commit that introduced the eval_model as it is not connected to the issue.

@pzhokhov I will have a look at the unit test later today...

PR #547, opened 2018-08-29
fix ppo2 tf.variable_scope reuse

studynon:
fix ppo2 `tf.variable_scope` reuse, in order to train ppo2's model in jupyter notebook.

### DIFF
--- a/baselines/ppo2/ppo2.py
+++ b/baselines/ppo2/ppo2.py
@@ -57,7 +57,8 @@ class Model(object):
             grads, _grad_norm = tf.clip_by_global_norm(grads, max_grad_norm)
         grads_and_var = list(zip(grads, var))
 
-        _train = trainer.apply_gradients(grads_and_var)
+        with tf.variable_scope('ppo2_model', reuse=tf.AUTO_REUSE):
+            _train = trainer.apply_gradients(grads_and_var)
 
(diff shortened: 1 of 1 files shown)


pzhokhov:
to my understanding, the problem is that re-running ppo2.learn with the same tensorflow graph tries to re-create trainable variables (including the state of the adam optimizer). While I agree that is a problem, I think we need a more principled solution - this problem will happen not only in ppo2; in most of the algorithms we use the same `trainer.compute_gradients` outside of explicit variable scope pattern. We'll either need to change all of them, or come up with some clever solution (maybe create scope within MpiAdamOptimizer).

pzhokhov:
@joschu @unixpickle any other ideas how can this be solved?

joschu:
I don't think it's a big problem, you can always create a new tf Graph.

studynon:
Thank you @pzhokhov @joschu , I have solved this issue via @joschu 's suggestion.

pzhokhov:
hm... my problem with such fix is that it bypasses the existing session configs. How about this - we can make make_session (baselines/common/tf_util.py) create new graph if the graph is not specified?

joschu:
That also sounds a bit risky, since make_session is used in a lot of places and someone might be relying on it using the default tf.Graph.

Issue #538, opened 2018-08-24
Gamma in VecNormalize for rms updates.


pisiiki:
I have observed that the VecNormalize class updates the reward running mean statistics with what looks to me like a discounted reward:

```
def step_wait(self):
    obs, rews, news, infos = self.venv.step_wait()
    self.ret = self.ret * self.gamma + rews # here
    obs = self._obfilt(obs)
    if self.ret_rms:
        self.ret_rms.update(self.ret)
        rews = np.clip(rews / np.sqrt(self.ret_rms.var + self.epsilon), -self.cliprew, self.cliprew)
    return obs, rews, news, infos
```

I can't see why this helps at all, I would directly use rews in the rms update. Am I missing something here?

Thanks.

pzhokhov:
it looks like the statistics are collected for episode returns (not for rewards), but I don't really know what's benefit of using that for the reward normalization vs using stats of the rewards themselves. Also, in that case, there may be a bug where you'd need to reset self.ret to 0 at the beginning of the episode, here: https://github.com/openai/baselines/blob/92b9a3725710a20ec62120e7b1533afc84a3df27/baselines/common/vec_env/vec_normalize.py#L39. @joschu @unixpickle could you comment please?

unixpickle:
Normalization is important because neural networks trained with Adam (and most other algorithms, as well) don't learn well with very large or very small targets. **Returns** are what we really want to normalize, because they are ultimately what drive the value function and policy.

To see why normalizing **returns** is different than normalizing **rewards**, take the simple case where the rewards are large but cancel each other out. For example, the rewards might be `[100, -101, 102, -99, ...]`. In this case, the returns will be fairly small, even though the rewards are large. Thus, the advantages we feed into the RL algorithm will already be fairly well-behaved. In a different environment, the rewards might be `[10, 10, 10, 10, ...]`. In this case, the rewards are not that large, but the returns are likely to be on the order of 1000 if we use gamma=0.99.

pzhokhov:
thanks @unixpickle! Do you agree with resetting returns to 0 in `reset` method, or is there some reason why we should not?

andytwigg:
I agree it makes sense to normalize returns. The current method effectively
applies a sliding window. I do have two concerns:

1) with gamma=0.99 this window is effectively of length ~100 which may be
less than a single episode, so the effect of normalizing across episodes is
lost.

2) we are dividing by std without subtracting mean. But the result is
clipped symmetrically around 0.

[...]

PR #556, opened 2018-08-31
VecNormalize: set env. returns to zero on resets.

pisiiki:
Related to this [issue](https://github.com/openai/baselines/issues/538).

Just in case, I guess this will help a bit.


andytwigg:
In line 24, news is the vector of dones, so I think you'd need to move this line to after ret_rms has been updated. 

Also perhaps a faster way would be to use news as a mask, ie after updating ret_rms, multiply ret by news. This would avoid the zip and creating a new ndarray.

pisiiki:
Thanks @andytwigg!

Should be fixed by now.

PR #560, opened 2018-08-31
refactor acktr

pzhokhov:


pzhokhov:
@tonywu95 could you please comment on whether such changes are reasonable? 
We are unable to get the same performance with refactored version as with acktr_cont. Unclear whether it is due to a bug in the refactored code, or due to missing weight decay / separate settings of kfac optimizer for value function / elu nonlinearities in the value function / something else. 
Also putting @joschu and @christopherhesse into the loop.
Thanks! 
acktr_cont.py (and related functions) will be available in a separate branch for experiments, `old_acktr_cont`.

pzhokhov:
Also, it looks like this implementation of acktr is basically a2c with KfacOptimizer (and separate learning rates for value function and policy). Is that so? Does it make sense to merge the code of the two?

PR #561, opened 2018-09-01
Error when dimension of stack differs

huvar:
stackedobs has the dimensions [stacks, env_obs, stacks]... In every step we should we be updating only the last stack, right? So we should update last instance of stack self.stackedobs[..., -1] = obs
Otherwise you get errors like:
ValueError: could not broadcast input array from shape (16,4) into shape (16,4,4)


pzhokhov:
could you provide an example of a failure? Ideally, as a test that fails before your change and passes after ^_^

PR #565, opened 2018-09-05
refactor common.models via registering reflection

studynon:
refactor common.models via registering reflection:

1. register default models into `get_network_builder` via `@register` decorator
2. make it more flexible to add new networks in the future
3. docs introducing how to register one's own network into `get_network_builder`


joschu:
LGTM

PR #571, opened 2018-09-07
fix return value of tf_util._check_shape

studynon:
fix `tf_util._check_shape` which returned hard-coded `True`


pzhokhov:
I feel like there was a corner case _check_shape was misbehaving... ah, there it is, shapes _check_shape([-1,2,3],[1,2,3]) returns False whereas it should be True. 
Feel free to change the logic of _check_shape to accomodate for that, but please make a unit test in that case.

PR #599, opened 2018-09-19
Update README.md

Linging:
common/cmd_util.py line 102: default value of num_env is None.
So if you run pong game under play mode, [--num_env = 4] is needed.


pzhokhov:
the command in README works for me - even though default value for num_env is None, run.py handles the case when it is None, here:
https://github.com/openai/baselines/blob/0e7048b89f93cd466dc52df4479e7488949daec1/baselines/run.py#L88
and here:
https://github.com/openai/baselines/blob/0e7048b89f93cd466dc52df4479e7488949daec1/baselines/run.py#L128

It looks like it is an issue for recurrent models, but let's rather fix that (the current line in README looks sensible to me, so if the code does not behave that way, let's fix the code :))

Issue #495, opened 2018-08-06
'async' is a reserved word in Python >= 3.7


cclauss:
flake8 testing of https://github.com/openai/baselines on Python 3.7.0

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./baselines/bench/monitor.py:142:55: F821 undefined name 'uuid'
    mon_file = "/tmp/baselines-test-%s.monitor.csv" % uuid.uuid4()
                                                      ^
./baselines/bench/monitor.py:158:20: F821 undefined name 'pandas'
    last_logline = pandas.read_csv(f, index_col=None)
                   ^
./baselines/bench/monitor.py:161:5: F821 undefined name 'os'
    os.remove(mon_file)    ^
./baselines/acktr/acktr_cont.py:57:69: E999 SyntaxError: invalid syntax
                                epsilon=1e-2, stats_decay=0.99, async=1, cold_iter=1,
                                                                    ^
./baselines/acktr/value_functions.py:24:41: E999 SyntaxError: invalid syntax
                                    async=1, kfac_update=2, cold_iter=50, \
                                        ^
./baselines/acktr/acktr_disc.py:62:39: E999 SyntaxError: invalid syntax
                stats_decay=0.99, async=1, cold_iter=10, max_grad_norm=max_grad_norm)
                                      ^
./baselines/acktr/kfac.py:13:164: E999 SyntaxError: invalid syntax
    def __init__(self, learning_rate=0.01, momentum=0.9, clip_kl=0.01, kfac_update=2, stats_accum_iter=60, full_stats_init=False, cold_iter=100, cold_lr=None, async=False, async_stats=False, epsilon=1e-2, stats_decay=0.95, blockdiag_bias=False, channel_fac=False, factored_damping=False, approxT2=False, use_float64=False, weight_decay_dict={},max_grad_norm=0.5):
                                                                                                                                                                   ^
4     E999 SyntaxError: invalid syntax
3     F821 undefined name 'uuid'
7
```

wookayin:
How about `is_async`? I think it would be a better name than `async_` (which is somewhat artificial).

cclauss:
I see other repos replacing __async__ with __non_blocking__.

cclauss:
Fixed in #542

PR #605, opened 2018-09-21
Travis CI: Upgrade to Python 3.7

cclauss:
Highlights the issue with #495 and #542

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
-python:
-    - "3.6"
+
+matrix:
+  include:
+    - python: 3.7
+      dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
+      sudo: required  # required for Python 3.7 (travis-ci/travis-ci#9069)
 
 services:
     - docker


cclauss:
You are misreading the Travis file.  All the test are NOT being run in a Docker container.

The flake8 tests are not being run inside docker.  Instead they are purposefully being run in the Python virtualenv created by Travis.  This is why the tests pass when the Travis virtualenv is Python 3.6 and fails when the Travis virtualenv is 3.7 even with an identical docker image.

pzhokhov:
Right, I think we have a naming convention mismatch; sorry if my statement was not clear. I agree that in flake8 (syntax, style etc) tests will be tested in both python3.6 and 3.7 in your PR. However, the unit tests are being run in python3.5 only, because the unit tests are run in the docker container; and I believe if we are making code 3.7 compatible, both style and unit tests need to run in both 3.6 and 3.7. That has, unfortunately, more strings attached - we need docker images with different versions python. To-date baselines code is loosely python3 compatible; adding explicitly flake8 on python3.6 and 3.7 for travis gives illusion of 3.7 compatibility (and does catch syntax errors) but is not a silver bullet. In terms of action plan - I would rather wait with such changes until we have unit tests running 3.6 and 3.7; other than that - looks good to me

PR #607, opened 2018-09-21
Wrote some comments to explain the A2C and PPO2 implementation

simoninithomas:
Hi,

1. I've added some comments in A2C implementations **to help the users to better understand this implementation. I tried to be succinct.**
--> I think it can help readers who try **to modify the implementations or really understand them.**

2. Also, I've modified the **readme file to explain rapidly what each file do.**

The new baseline version **is really good**. However I think we can improve readme part, or what if include a wiki part **to describe the different architectures with schemas, because it can helps users to have the big picture in mind** (what file does what, what is the process etc). [Like the schema I made for the former PPO2 version](http://bit.ly/2DsbsSe) but not as ugly as mine version 😂 (this is a draft).

### DIFF
--- a/baselines/a2c/README.md
+++ b/baselines/a2c/README.md
@@ -4,3 +4,10 @@
 - Baselines blog post: https://blog.openai.com/baselines-acktr-a2c/
 - `python -m baselines.run --alg=a2c --env=PongNoFrameskip-v4` runs the algorithm for 40M frames = 10M timesteps on an Atari Pong. See help (`-h`) for more options
 - also refer to the repo-wide [README.md](../../README.md#training-models)
+
+## Files
+- `run_atari`: file used to run the algorithm.
+- `policies.py`: contains the different versions of the A2C architecture (MlpPolicy, CNNPolicy, LstmPolicy...).
+- `a2c.py`: - Model : class used to initialize the step_model (sampling) and train_model (training)
+	- learn : Main entrypoint for A2C algorithm. Train a policy with given network architecture on a given environment using a2c algorithm.
+- `runner.py`: class used to generates a batch of experiences
--- a/baselines/a2c/a2c.py
+++ b/baselines/a2c/a2c.py
@@ -21,10 +21,10 @@ class Model(object):
         __init__:
         - Creates the step_model
         - Creates the train_model
-        
+
         train():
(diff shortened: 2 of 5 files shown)


pzhokhov:
Thanks @simoninithomas, I really like this! Could you also autopep it (`autopep8 -i -aa -r .`) to fix the flake8 complaints? 
Regarding diagrams to explain how things work - sounds like a great idea! Putting @jachiam and @joschu in the loop, they may have opinion on how to explain things best.

simoninithomas:
Awesome, thanks @pzhokhov !

simoninithomas:
Hi @jachiam , thanks for the feedback, I'm updating my pull request

I create a new PR because this one is merged 

========
a2c/a2c.py

L55: **Modified**

L86: **Modified**

L93: " Strongly prefer that a comment describe the literal calcualtion performed by the code, instead of the ideal case." Ok so if I understood, you consider that writing rewards like = R + yV(s') (or r_t + y * V(s')) can mislead others (because of the fact that's the best case scenario aka when we have a "perfect" critic).

But I'm not sure about if this what you want (hoping one day github will implement LaTeX) ?
`# rewards =  k-step rewards + y * predicted_value(s_t+k+1)`

L218-L229: Indeed, I thought it was easier for users to find the explanation directly in the main file. But we can reduce it to

`Calculates if value function is a good predicator of the returns (ev >1) or if it's just worse than predicting nothing (ev=<0)`
? 

==========
a2c/runner.py

L27: **Modified**

===============
common/policies.py

L50: My bad it was initially commented to explain the PPO with Sonic, I remove that comment.

==========
ppo2/ppo2.py

Would remove them and maybe write a comment to justify why the value clipping is done at all : **Modified** I suppose we clip the value to add stability of the Critic training by limiting the value update at each training step.
So I wrote `# Clip the value to reduce variability during Critic training`

Too many lines of comments are devoted to something akin to mathematical derivation: Indeed, but it was to help users who wants to really understand the mathematical logic behind. But I'm agree that's not relevant in an implementation. **I remove it**. But it can be interesting to add it maybe in a wiki on something like that in the future because this part of the code is not intuitive (not because of the way is coded but because of the mathematical manipulations we need to do).

L187: **Deleted**

Comments on GAE: **Deleted**

L394-401: **Modified**

PR #609, opened 2018-09-21
Adding num_env to readme example

ahuhn:
There is a bug in the current README file that causes this example to throw an error. The error thrown by

`python -m baselines.run --alg=ppo2 --env=PongNoFrameskip-v4 --num_timesteps=0 --load_path=~/models/pong_20M_ppo2 --play`

is

`[...]"/openai/baselines/baselines/run.py", line 227, in initialize_placeholders
    return np.zeros((args.num_env, 2*nlstm)), np.zeros((1))
TypeError: 'NoneType' object cannot be interpreted as an integer`

I added `--num_env=4` to the example, and I also fixed the underlying bug where it assumes that args.num_env is an integer. (I'm not sure I used best-practices for that change, so feel free to suggest a different fix.)

### DIFF
--- a/baselines/run.py
+++ b/baselines/run.py
@@ -222,7 +222,7 @@ def main():
         env = build_env(args)
         obs = env.reset()
         def initialize_placeholders(nlstm=128,**kwargs):
-            return np.zeros((args.num_env, 2*nlstm)), np.zeros((1))
+            return np.zeros((args.num_env or 0, 2*nlstm)), np.zeros((1))
         state, dones = initialize_placeholders(**extra_args)
         while True:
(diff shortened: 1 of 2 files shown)


baselines/run.py
```diff
@@ -222,7 +222,7 @@ def main():
         env = build_env(args)
         obs = env.reset()
         def initialize_placeholders(nlstm=128,**kwargs):
-            return np.zeros((args.num_env, 2*nlstm)), np.zeros((1))
+            return np.zeros((args.num_env or 0, 2*nlstm)), np.zeros((1))
```
pzhokhov: also, I think this should be "or 1" instead of or 0

PR #632, opened 2018-10-03
fix DQN learning bug

pzhokhov:
the problem was input was scaled twice - once by the ScaledFloatFrame Wrapper, once by the neural net code.

### DIFF
--- a/baselines/deepq/utils.py
+++ b/baselines/deepq/utils.py
@@ -1,8 +1,6 @@
 from baselines.common.input import observation_input
 from baselines.common.tf_util import adjust_shape
 
-import tensorflow as tf
-
 # ================================================================
 # Placeholders
 # ================================================================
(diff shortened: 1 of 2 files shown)


wookayin:
How about explicitly stating as `scale=False`? I don't think the default argument value of `wrap_deepmind()` will change anytime soon, but it is more explicit and clear as rescaling is done by the DQN's nerual network code.

pzhokhov:
@christopherhesse the second scaling was happening in the input of cnn (baselines/common/models.py:conv_only)
@wookayin Good point, however, I'd rather eventually get rid of the split logic in run.py (where different set of wrappers is built for different algorithms

christopherhesse:
Ah, thanks.  I also saw some scaling here: https://github.com/openai/baselines/blob/4e2a888273f14c437fe6e036904e91c122ff9643/baselines/deepq/utils.py#L43 but this function doesn't appear to be used anywhere.

pzhokhov:
yes; in fact, let me delete it

pzhokhov:
yes, good point, thanks!

PR #654, opened 2018-10-16
sync internal changes. Make ddpg work with vecenvs

pzhokhov:


baselines/ddpg/ddpg.py
```diff
@@ -133,21 +145,24 @@ def learn(network, env,
...
-                    # Episode done.
-                    epoch_episode_rewards.append(episode_reward)
-                    episode_rewards_history.append(episode_reward)
-                    epoch_episode_steps.append(episode_step)
-                    episode_reward = 0.
-                    episode_step = 0
-                    epoch_episodes += 1
-                    episodes += 1
+                for d in range(len(done)):
+                    if done[d] == True:
```
joschu: ==True unnecessary

baselines/ddpg/memory.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, limit, action_shape, observation_shape):
 
     def sample(self, batch_size):
         # Draw such that we always have a proceeding element.
-        batch_idxs = np.random.random_integers(self.nb_entries - 2, size=batch_size)
+        batch_idxs = np.random.randint(self.nb_entries - 2, size=batch_size)
```
araffin: random_integers and randint are not equivalent, the interval is slightly different... see https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/ddpg/memory.py#L91

pzhokhov: that's a fair point; but I don't think there is a reason to exclude element 0... I'll test it to double-check this does not degrade the performance, but it would be really surprizing if it does.

Issue #651, opened 2018-10-14
Scope issues


manuel-delverme:
I'm trying to load several models in the same run but the TF scope is being reused
adding a ```with tf.variable_scope(...)``` was not enough since some of the scopes are hardcoded

this code is enough to crash, i would like to keep both models in memory.
``` 
model1 = baselines.trpo_mpi.trpo_mpi.learn(network='mlp', env=env, total_timesteps=0,)
model2 = baselines.trpo_mpi.trpo_mpi.learn(network='mlp', env=env, total_timesteps=0,)
```
how can i achieve that without modifying the baselines code?

manuel-delverme:
#547 suggests using a new graph but I need both models alive.

DanielTakeshi:
@manuel-delverme I handle loading models in the same run by making different scopes when the models are formed during training time (so basically, it's about ensuring that training is done by wrapping the entire model creation inside a tf scope).

pzhokhov:
Yeah, I could not find the way of doing it in the current implementation. However, with small changes (https://github.com/openai/baselines/pull/655) this can be done as in the follows:

```
import tensorflow as tf
import gym
from baselines.trpo_mpi.trpo_mpi import learn
from baselines.common.tf_util import make_session

env = gym.make('CartPole-v0')
make_session(make_default=True, graph=tf.Graph()); model1 = learn(network='mlp', total_timesteps=0, env=env)
make_session(make_default=True, graph=tf.Graph()); model2 = learn(network='mlp', total_timesteps=0, env=env)

# both models are alive: 
print(model1.step(env.observation_space.sample())
print(model2.step(env.observation_space.sample())
```

pzhokhov:
tested for a2c, acktr, ppo2, trpo_mpi; this will not work for deepq though

pzhokhov:
closing this (but opened a separate issue to enable multiple DQN / DDPG models to be present at the same time: https://github.com/openai/baselines/issues/656)

PR #655, opened 2018-10-17
store session at policy creation time

pzhokhov:
this way, multiple policies of the same kind (even the ones that were created with the default sessions) may coexist. Related to issue https://github.com/openai/baselines/issues/651

### DIFF
--- a/baselines/common/tests/envs/mnist_env.py
+++ b/baselines/common/tests/envs/mnist_env.py
@@ -1,7 +1,6 @@
 import os.path as osp
 import numpy as np
 import tempfile
-import filelock
 from gym import Env
 from gym.spaces import Discrete, Box
 
@@ -14,6 +13,7 @@ class MnistEnv(Env):
             episode_len=None,
             no_images=None
     ):
+        import filelock
         from tensorflow.examples.tutorials.mnist import input_data
         # we could use temporary directory for this with a context manager and
         # TemporaryDirecotry, but then each test that uses mnist would re-download the data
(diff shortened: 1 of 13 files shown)


baselines/common/tests/envs/mnist_env.py
```diff
@@ -14,6 +13,7 @@ def __init__(
             episode_len=None,
             no_images=None
     ):
+        import filelock
```
christopherhesse: any reason for the late imports here?

baselines/common/tests/test_serialization.py
```diff
@@ -75,6 +76,41 @@ def test_serialization(learn_fn, network_fn):
...
+
+    if learn_fn == 'deepq':
+            # TODO enable multiple DQN models to be useable at the same time
+            # github issue https://github.com/openai/baselines/issues/656
+            return
+
+    if network_fn.endswith('lstm') and learn_fn in ['acktr', 'trpo_mpi', 'deepq']:
+            # TODO make acktr work with recurrent policies
+            # and test
+            # github issue: https://github.com/openai/baselines/issues/194
```
christopherhesse: I like that these TODOs have links to the corresponding issues, the second link doesn't work for me though.

baselines/common/tests/envs/mnist_env.py
pzhokhov: so that we don't need filelock module for the tests that don't use MnistEnv - not a really big reason, but sometimes comes handy.

pzhokhov:
@christopherhesse ideally - yes, separate scopes should be sufficient; however, many algorithms at the moment use something like `tf.trainable_variabes()` to find variables to be updated which does not respect scopes.

PR #661, opened 2018-10-18
disable gym subpackages in setup.py

pzhokhov:
the installation of latest mujoco-py fails if mujoco is not present; so baselines that requires gym[mujoco] that requires mujoco-py fails as well.  On the other hand, we don't actually need that for baselines to work


pzhokhov:
dunno, actually... I guess quirk of how subpackages are processed in extras section?

PR #664, opened 2018-10-20
refactor ACER

pzhokhov:
make ACER use VecFrameStack wrapper instead of internal stacker, 
make it compatible with non-image observations, 
cartpole test


pzhokhov:
benchmarked on atari10M - gets the same learning curves as before this mini-refactor

PR #665, opened 2018-10-20
minor fixes from internal

pzhokhov:


pzhokhov:
good question. I guess the right way of doing it would be create a respective gym issue... @joschu could you create a github issue about that please?

joschu:
actually this is an issue with the rcall image or games installation -- currently we install an old version of gym

joschu:
the issue is already fixed in gym

christopherhesse:
Should we update gym in games @pzhokhov?

PR #666, opened 2018-10-20
Add video recorder

coord-e:
I added `VecVideoRecorder` to save mp4 video from `VecEnv`, and added `--video_monitor` and `--video_length` flags to use it.
Users can configure the interval and length of video recording via these flags.

Recorded videos will saved in `videos` directory under logger's path.


baselines/common/vec_env/vec_video_recorder.py
```diff
@@ -0,0 +1,70 @@
...
+        self.recorded_frames = 1
+        self.recording = True
+
+    def _video_enabled(self):
+        return self.video_callable(self.episode_id)
+
+    def step_wait(self):
+        obs, rews, dones, infos = self.venv.step_wait()
+
+        self.episode_id += 1
```
pzhokhov: if we аre increasing self.episode_id each step, would not be more correct to call it step_id?

baselines/common/vec_env/vec_video_recorder.py
```diff
@@ -0,0 +1,70 @@
+import os
+from baselines import logger
+from baselines.common.vec_env import VecEnvWrapper
+from gym.wrappers.monitoring import video_recorder
+
+
+class VecVideoRecorder(VecEnvWrapper):
+    def __init__(self, venv, directory, video_callable, video_length=200):
+        VecEnvWrapper.__init__(self, venv)
+        self.video_callable = video_callable
```
pzhokhov: the role of video_callable parameter is not very clear to me... Looks like it is something that defines when to start recording the video, right? Either a more transparent name (maybe start_recording_trigger?)  or a docstring, better yet both :) would be great.

baselines/common/cmd_util.py
```diff
@@ -102,6 +102,8 @@ def common_arg_parser():
     parser.add_argument('--num_env', help='Number of environment copies being run in parallel. When not specified, set to number of cpus for Atari, and to 1 for Mujoco', default=None, type=int)
     parser.add_argument('--reward_scale', help='Reward scale factor. Default: 1.0', default=1.0, type=float)
     parser.add_argument('--save_path', help='Path to save trained model to', default=None, type=str)
+    parser.add_argument('--video_monitor', help='Save video every x steps (0 = disabled)', default=0, type=int)
```
pzhokhov: I am not a fan of this argument name... to me, that looks more like a flag that determines if video_monitor is turned on or not. Maybe "save_video_interval"?

coord-e:
@pzhokhov Thank you for reviewing.
I pushed the changes including naming fixes and a test for `VecVideoRecorder`.

baselines/common/vec_env/vec_video_recorder.py
coord-e: Actually, I took this name from code in Gym's wrapper.

https://github.com/openai/gym/blob/520fd3a5e3e1362cf1ab4c03afe90e4e07d47007/gym/wrappers/monitor.py#L13

I added docstring of arguments in [this commit](https://github.com/openai/baselines/pull/666/commits/dd62945fe08acbedebc9c12df589372d1b4cb7a1).
If I should change the name of this argument, please tell me.

pzhokhov: we are between a rock and a hard place then... I'd say let's change the name to `record_video_trigger` nonetheless. In gym the name is also not particularly illuminating, and also works slightly differently from what I understand (takes number of the episode and returns if it should be recorded or not, whereas here it takes number of step, and returns whether we should start recording or not). Otherwise, LGTM

coord-e: Thank you for suggestion! fixed in https://github.com/openai/baselines/pull/666/commits/99f9b8e14fbcf720596ec7a2c82e84bbd3970791

Issue #646, opened 2018-10-11
AttributeError: 'ScaledFloatFrame' object has no attribute 'num_envs'


MatPoliquin:
When I run:
python3 -m baselines.run --alg=ppo2 --env=SuperMarioBros-Nes --num_timesteps=2e7

I get this error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/fanbingbing/ML/baselines/baselines/run.py", line 239, in <module>
    main()
  File "/home/fanbingbing/ML/baselines/baselines/run.py", line 213, in main
    model, env = train(args, extra_args)
  File "/home/fanbingbing/ML/baselines/baselines/run.py", line 79, in train
    **alg_kwargs
  File "/home/fanbingbing/ML/baselines/baselines/ppo2/ppo2.py", line 295, in learn
    nenvs = env.num_envs
AttributeError: 'ScaledFloatFrame' object has no attribute 'num_envs'
```

I made a temporary fix but get other similar errors further.

Additional info:
Ubuntu 18.04
Python 3.6.6
synced baselines with latest commit: 28aca63

retro env works fine with SuperMario when I test it directly so it's not an issue with retro env not working
Also the example command provided in the readme works fine:
python3 -m baselines.run --alg=ppo2 --env=PongNoFrameskip-v4 --num_timesteps=2e7

pzhokhov:
confirmed, will fix shortly

MatPoliquin:
@pzhokhov Thanks a lot!

pzhokhov:
should be fixed now

MatPoliquin:
After syncing the latest commit and running the exact same command and setup as described above I get this error now:

Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/fanbingbing/ML/baselines/baselines/run.py", line 221, in <module>
    main()
  File "/home/fanbingbing/ML/baselines/baselines/run.py", line 195, in main
    model, env = train(args, extra_args)
  File "/home/fanbingbing/ML/baselines/baselines/run.py", line 78, in train
    **alg_kwargs
[...]

PR #669, opened 2018-10-22
wrap retro envs correctly for other (non-deepq) algorithms

pzhokhov:
this enables, for instance, ppo2 on retro games (related to this: https://github.com/openai/baselines/issues/646)

### DIFF
--- a/baselines/common/cmd_util.py
+++ b/baselines/common/cmd_util.py
@@ -26,7 +26,7 @@ def make_vec_env(env_id, env_type, num_env, seed, wrapper_kwargs=None, start_ind
     mpi_rank = MPI.COMM_WORLD.Get_rank() if MPI else 0
     seed = seed + 10000 * mpi_rank if seed is not None else None
     def make_thunk(rank):
-        return lambda: env_thunk(
+        return lambda: make_env(
             env_id=env_id,
             env_type=env_type,
             subrank = rank,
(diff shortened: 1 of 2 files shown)


baselines/common/cmd_util.py
```diff
@@ -16,30 +16,58 @@
...
     set_global_seeds(seed)
-    if num_env > 1: return SubprocVecEnv([make_env(i + start_index) for i in range(num_env)])
-    else: return DummyVecEnv([make_env(start_index)])
+    if num_env > 1:
+        return SubprocVecEnv([make_thunk(i + start_index) for i in range(num_env)])
+    else:
+        return DummyVecEnv([make_thunk(start_index)])
+
+
+def env_thunk(env_id, env_type, subrank=0, seed=None, reward_scale=1.0, gamestate=None, wrapper_kwargs=None):
```
christopherhesse: should this be make_env to correspond to make_vec_env above?

baselines/common/cmd_util.py
```diff
@@ -16,30 +16,58 @@
 from baselines.common.atari_wrappers import make_atari, wrap_deepmind
 from baselines.common.vec_env.subproc_vec_env import SubprocVecEnv
 from baselines.common.vec_env.dummy_vec_env import DummyVecEnv
-from baselines.common.retro_wrappers import RewardScaler
+from baselines.common import retro_wrappers
+import functools
```
christopherhesse: I think this isn't used anywhere

pzhokhov: yes, thanks! flake8 flags it now (was delayed due to github issues)

Omegastick:
I think this broke `make_vec_env` for Atari environments. `wrapper_kwargs` isn't passed in `make_thunk` so it is always `None`. This gives `TypeError: wrap_deepmind() argument after ** must be a mapping, not NoneType` when you call it. I'll open an issue.

Issue #670, opened 2018-10-23
TypeError: wrap_deepmind() argument after ** must be a mapping, not NoneType


Omegastick:
I think #669 broke make_vec_env for Atari environments. `wrapper_kwargs` isn't passed in `make_thunk` so it is always `None`. This gives `TypeError: wrap_deepmind() argument after ** must be a mapping, not NoneType` when you call it.

I've got a fix ready on my machine if you guys would like a pull request.

PR #671, opened 2018-10-23
Fixed TypeError on creating atari vec envs (introduced in #669)

Omegastick:
As detailed in #670, there was a bug in #669. Simple fix.


christopherhesse:
@pzhokhov should the tests cover this code path?

PR #674, opened 2018-10-23
algorithm registry

pzhokhov:


christopherhesse:
I like that everything is imported in __init__.py so that things are registered, but is this better than having, say, the import path be the registered name?  Like if instead of code to do registration, I just have data like the train function is 'baselines.a2c.a2c.learn' and here are the default values for different environments in a dict.

pzhokhov:
@joschu regarding the defaults - the idea is that the "defaults" argument are defaults fine-tuned to a specific kind of environment (atari, retro etc). The values that don't change do not need to be repeated; so I don't think reading default values from the function would help much with that. 

@christopherhesse the main improvement over the import path convention is that with registry we can have extra metadata (for instance, if algorithm supports vecenvs or not; maybe in the future what types of observation / action spaces and policies is algorithm compatible with etc. 
That being said, we can do it via import paths too (by making algorithm have "metadata" module). 
The one advantage that registry have that you cannot get through standard import paths is adding external algorithms (like we have addition of new environments to the gym). But I am not sure it is all that necessary.

pzhokhov:
@joschu the example of using the registry would be something like this:
```
from baselines import registry
@registry.register('fancy_algorithm', supports_vecenv=False)
def learn(env, network):
  # some fancy algorithm

for algo_name, algo_entry in registry.registry.items():
  if not algo_entry['supports_vecenv']:
     print(f'{algo_name} does not support vecenvs
     # should print "fancy_algorithm does not support vecenvs" (among other ones)"
```

I'll write a comment to that extent in baselines.registry module

christopherhesse:
With the import path version, adding a new algorithm would be a matter of having some relevant metadata and providing the import path to whatever function is going to run this right?  I'm totally fine with the registry, I just think it's a little awkward in python because you end up having to import things for their registration side effects.

pzhokhov:
yeah, the two registries can be confusing... you know what, let me close this for know and take a stab at import path based approach with metadata (we already had most of that before, so it should not be that big of a deal)

PR #680, opened 2018-10-25
DDPG: noise_type 'normal_x' and 'ou_x' cause AssertionError

AurelianTactics:
noise_type default 'adaptive-param_0.2' works but the arguments that change from parameter noise to actor noise (like 'normal_0.2' and 'ou_0.2') cause an assert message and DDPG not to run. Issue is the following block:
```
        if self.action_noise is not None and apply_noise:
            noise = self.action_noise()
            assert noise.shape == action.shape
            action += noise
```

noise is not nested: [number_of_actions]
action is nested: [[number_of_actions]]
Hence the shapes do not pass the assert line even though the action += noise line is correct

### DIFF
--- a/baselines/acer/acer.py
+++ b/baselines/acer/acer.py
@@ -7,6 +7,7 @@ from baselines import logger
 from baselines.common import set_global_seeds
 from baselines.common.policies import build_policy
 from baselines.common.tf_util import get_session, save_variables
+from baselines.common.vec_env.vec_frame_stack import VecFrameStack
 
 from baselines.a2c.utils import batch_to_seq, seq_to_batch
 from baselines.a2c.utils import cat_entropy_softmax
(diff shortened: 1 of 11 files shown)


pzhokhov:
LGTM, but please add a test that would fail without this change and pass with it (maybe instantiate DDPG with different noise types

AurelianTactics:
Below are some tests that fail under current code but work with pull request. Any use of action noise either on its own or chained together with adaptive parameter noise (third test) causes the assertion error.

python -m baselines.run --alg=ddpg --env=Pendulum-v0 --noise_type=normal_0.1
python -m baselines.run --alg=ddpg --env=Pendulum-v0 --noise_type=ou_0.1
python -m baselines.run --alg=ddpg --env=Pendulum-v0 --noise_type=adaptive-param_0.2,normal_0.1

The error:

> Traceback (most recent call last):
> ...
>   File ".../baselines/baselines/run.py", line 221, in <module>
>     main()
>   File ".../baselines/baselines/run.py", line 195, in main
>     model, env = train(args, extra_args)
>   File ".../baselines/baselines/run.py", line 78, in train
>     **alg_kwargs
>   File "../baselines/baselines/ddpg/ddpg.py", line 129, in learn
>     action, q, _, _ = agent.step(obs, apply_noise=True, compute_Q=True)
>   File ".../baselines/baselines/ddpg/ddpg_learner.py", line 271, in step
>     assert noise.shape == action.shape
> AssertionError

pzhokhov:
I meant unit tests so that travis can run them, but this is good enough - I'll add calling baselines.run from python and then make these cases unit test cases.

PR #682, opened 2018-10-27
Add retro to ppo2 defaults

MatPoliquin:
Created defaults for retro, copied from Atari defaults for now. Tested with SuperMarioBros-Nes

### DIFF
--- a/baselines/ppo2/defaults.py
+++ b/baselines/ppo2/defaults.py
@@ -22,10 +22,4 @@ def atari():
     )
 
 def retro():
-    return dict(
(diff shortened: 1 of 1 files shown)


baselines/ppo2/defaults.py
```diff
@@ -20,3 +20,12 @@ def atari():
         lr=lambda f : f * 2.5e-4,
         cliprange=lambda f : f * 0.1,
     )
+
+def retro():
```
pzhokhov: I'd say more concise way here would be to do
```
def retro():
     return atari()
```

PR #684, opened 2018-10-27
README in french

BansalHere:


pzhokhov:
uhm... why do we need to translate the documentation into French?

PR #685, opened 2018-10-27
Use deepmind wrapper for retro

MatPoliquin:


baselines/common/cmd_util.py
```diff
@@ -51,6 +51,7 @@ def make_env(env_id, env_type, subrank=0, seed=None, reward_scale=1.0, gamestate
         import retro
         gamestate = gamestate or retro.State.DEFAULT
         env = retro_wrappers.make_retro(game=env_id, max_episode_steps=10000, use_restricted_actions=retro.Actions.DISCRETE, state=gamestate)
+        env = retro_wrappers.wrap_deepmind_retro(env)
```
pzhokhov: I think this needs to be after Monitor; otherwise, Monitor will report modified rewards. Also, we can pass `wrapper_kwargs` to `wrap_deepmind_retro`.

MatPoliquin:
Moved Deepmind retro wrapper after monitor.
As for RewardScaler it was only possibly called for non-atari env, was it intended this way?

pzhokhov:
atari usually come with reward clip wrapper, so reward scaler does not help much there; but I don't see harm in allowing it in principle.

PR #689, opened 2018-10-29
mpi-less baselines

pzhokhov:
make baselines work on systems without MPI


pzhokhov:
less of a hassle during installation. One example that I ran into is trying it on colab notebooks (wanted to add some working examples in the docs). While it is possible to install mpi on colab, I'd rather keep examples minimal.

PR #694, opened 2018-10-31
Internal minifixes

pzhokhov:


araffin:
Hello,
The current failure on travis is related to https://github.com/openai/gym/issues/1191 ;)

pzhokhov:
yes, I agree. I think the right thing to do here would be to fix the MultiDiscrete constructor in the gym (I don't think there is a reason to wrap tuple in another tuple)

PR #703, opened 2018-11-06
more viz + build fixes

pzhokhov:


.travis.yml
```diff
@@ -11,4 +11,4 @@ install:
 
 script:
     - flake8 . --show-source --statistics
-    - docker run baselines-test pytest -v .
```
christopherhesse: is forked better than using xdist?

pzhokhov: No, but I think travis do not have enough cores to actually use parallel version of pytest; here forked is needed rather for test isolation (the new video recorder tests freeze without it)

christopherhesse: Ah, my only concern is that pytest-forked has a warning about maintenance: https://github.com/pytest-dev/pytest-forked#pytest-forked-run-each-test-in-a-forked-subprocess while xdist has no such warning.

pzhokhov:
that's a good point! I did not think of putting the entire doc into the colab notebook, but we can totally do that.

christopherhesse:
@pzhokhov 

Related to that, github will just display ipynb files: https://github.com/norvig/pytudes/blob/master/ipynb/BASIC.ipynb

So having it all in a notebook that is also part of the repo should be doable, you can link to the colab version like so: https://colab.research.google.com/github/norvig/pytudes/blob/master/ipynb/BASIC.ipynb

PR #715, opened 2018-11-15
[VecEnv]: add action clipping wrapper

zuoxingdong:
Hi @pzhokhov , do you think this wrapper can be useful ?

This might be very convenient for certain continuous control environments where the policy network might sample action outside the bounds. And it avoids runtime error for the users without writing their own clipping. 

Another advantage is that in policy gradient methods, sometimes people make mistake to calculate log-probability of sampled action after doing clipping, this will in fact transform the underlying distributions and may degrade the performance. By providing this wrapper, it tries to avoid this mistake from users side.

### DIFF
--- a/baselines/common/vec_env/vec_action_clip.py
+++ b/baselines/common/vec_env/vec_action_clip.py
@@ -1,0 +1,17 @@
+from . import VecEnvWrapper
+import numpy as np
+
+
+class VecActionClip(VecEnvWrapper):
+    def step_async(self, actions):
+        low = self.action_space.low
+        high = self.action_space.high
+        actions = [np.clip(action, low, high) for action in actions]
+        
+        self.venv.step_async(actions)
+    
+    def step_wait(self):
+        return self.venv.step_wait()
+    
+    def reset(self):
+        return self.venv.reset()


pzhokhov:
I like this, but please add unit tests

andytwigg:
Wouldn't it make more sense to have this clipping enforced in spaces.Box
directly?

zuoxingdong:
@andytwigg If I got it correctly, do you mean to add a function to `Box` class for clipping ? If so, I guess it is still better to deal with it directly with VecEnv wrapper. Because it has something to do with the actions that environment step function receives from the users, often produced by neural network policies, in which having invalid actions out of the bound is possible.

andytwigg:
It would seem intuitive that for some space eg Box(0,1,(10,)):

pdtype = make_pdtype(space)
pd, pi = pdtype.pdfromlatent(h)
assert space.contains(pd.sample())

which of course if not the case unless you manually clip. Adding clipping
to vecenv wouldn't fix this; it would only work for vecenvs?

zuoxingdong:
@andytwigg Ahaa, but then it will raise runtime error when action goes out of the bound. When training RL agent, this is probably not the behavior we want to have.

andytwigg:
That's right, which is why i suggested putting tf.clip_by_value(low,high)
into pdfromlatent directly?

pzhokhov:
@andytwigg you mean make_pdtype create gaussian distribution that is clipped (instead of simply gaussian) when action space limits are finite? That's a good idea, but certain care has to be taken when neglogp and entropy of such distribution.

pzhokhov:
To state my position more clearly - @andytwigg's proposed solution is more principled and elegant (I would also add beta distribution in that case that inherently produces samples in limited range), but a bit more complex, whereas @zuoxingdong's one is more of a hack, but I am ok with it being merged if outside of action limits causes errors. Also, a random thought - it does not have to be on a vecenv level, it could be an individual sub-environment wrapper.

andytwigg:
good point about neglogp+entropy. How about adding a method like
b=space.ensure(a) which will ensure that space.contains(b) is True? That
way, the clipping logic etc can be in the same place as the space itself. I
guess having the clip logic living separately is what worried me.

pzhokhov:
oops sorry about the lag. I don't think is matters much where does the actual clipping logic lives - even if the itself has a function `ensure` it would still be up to user to call / not call that function... All in all, I am inclined to merge @zuoxingdong's wrapper till we are ready to have more principled solution.

baselines/common/vec_env/test_vec_clip_action.py
```diff
@@ -0,0 +1,28 @@
...
+    env = klass(fns)
+    env_clipped = VecClipAction(env)
+
+    env.reset()
+    env_clipped.reset()
+
+    action = [[0.5], [1000]]
+
+    _, rewards, _, _ = env.step(action)
+    _, rewards_clipped, _, _ = env_clipped.step(action)
```
pzhokhov: these are two steps on the same environment (instead of the copy) - is that intentional?

zuoxingdong:
@pzhokhov The only question I am thinking about is if this wrapper could be fitting more naturally in gym repo or at here.

PR #733, opened 2018-11-26
fixed issue with wrong output layer variable names in ddpg

pzhokhov:


joschu:
LGTM

baselines/ddpg/ddpg_learner.py
```diff
@@ -185,7 +185,7 @@ def setup_critic_optimizer(self):
         normalized_critic_target_tf = tf.clip_by_value(normalize(self.critic_target, self.ret_rms), self.return_range[0], self.return_range[1])
         self.critic_loss = tf.reduce_mean(tf.square(self.normalized_critic_tf - normalized_critic_target_tf))
         if self.critic_l2_reg > 0.:
-            critic_reg_vars = [var for var in self.critic.trainable_vars if 'kernel' in var.name and 'output' not in var.name]
+            critic_reg_vars = [var for var in self.critic.trainable_vars if var.name.endswith('/w:0') and 'output' not in var.name]
```
Lukeeeeee: Hi, may I know the reason why use `/w:0` instead of `kernel`  to filter the tf variable? 
Since you use `tf.layers.dense` in critic network, while the implementation is here: https://github.com/tensorflow/tensorflow/blob/a6d8ffae097d0132989ae4688d224121ec6d8f35/tensorflow/python/keras/layers/core.py#L937

And the source code shows that it will append the name 'kernel' instead of 'w:0'.

pzhokhov: variables in question here (the ones that will be regularized) are all but the last layer; and those are built not using tf.layers.dense, but rather using baselines.a2c.utils.fc - see here:
https://github.com/openai/baselines/blob/57e05eb420f9a20fa8cd7ee7580b1a8874b4a323/baselines/ddpg/models.py#L44, https://github.com/openai/baselines/blob/57e05eb420f9a20fa8cd7ee7580b1a8874b4a323/baselines/common/models.py#L52, and https://github.com/openai/baselines/blob/57e05eb420f9a20fa8cd7ee7580b1a8874b4a323/baselines/a2c/utils.py#L58.
There reason for that is more historical (and eventually we'll switch them to tf.dense), but for now the names of weights of inner layers (at least of the default networks) end with '/w:0'.

Issue #739, opened 2018-11-29
Unable to debug the PPO2 algorithm with VSCode


davidhopper2003:
I used VSCode to debug the PPO2 algorithm. When debugging to line 58 of the file `baselines/common/vec_env/subproc_vec_env.py`:

```python
observation_space, action_space = self.remotes[0].recv()
```

the program has no response and the debugging process cannot continue (As shown below). 
![ppo2_debug_1](https://user-images.githubusercontent.com/31435476/49197695-c8a4ff80-f3ca-11e8-87f0-468bb47250c2.png)
![ppo2_debug_2](https://user-images.githubusercontent.com/31435476/49197696-c8a4ff80-f3ca-11e8-84a0-6297ab5f8868.png)
![ppo2_debug_3](https://user-images.githubusercontent.com/31435476/49197697-c8a4ff80-f3ca-11e8-989e-2ecbba52c6b1.png)

My configuration file `.vscode/launch.json` is shown as follows:

```
{

    // Use IntelliSense to learn about possible attributes.

    // Hover to view descriptions of existing attributes.

    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387

    "version": "0.2.0",

    "configurations": [

        {

            "name": "Python: Module",

            "type": "python",

            "request": "launch",

            "module": "baselines.run",

            "console": "integratedTerminal",

            "args": [

                "--alg",

                "ppo2",

                "--env",

                "PongNoFrameskip-v4",

                "--num_timesteps",

                "2e7",

                "--save_path",

                "~/models/pong_20M_ppo2"

            ],

        },

        {

            "name": "Python: Current File (Integrated Terminal)",

            "type": "python",

            "request": "launch",

            "program": "${workspaceFolder}/baselines/run.py",

            "console": "integratedTerminal", 

            "args": [

                "--alg",

                "ppo2",

                "--env",

                "PongNoFrameskip-v4",

                "--num_timesteps",

                "2e7",

                "--save_path",

                "~/models/pong_20M_ppo2"

            ],

        }

    ]

}
```

When I used the command line 

`python -m baselines.run --alg=ppo2 --env=PongNoFrameskip-v4 --num_timesteps=2e7 --save_path=~/models/pong_20M_ppo2` 

to run the program in a terminal, there is no problem. 

Any helpful suggestions will be highly appreciated. Thank you.

davidhopper2003:
[...]

PR #745, opened 2018-11-30
Added support for debugging multi-process tasks in VSCode.

davidhopper2003:
Refer to my issue [Unable to debug the PPO2 algorithm with VSCode](https://github.com/openai/baselines/issues/739).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,84 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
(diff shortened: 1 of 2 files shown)


ashrafulislamM:
Thanks

.vscode/launch.json
```diff
@@ -0,0 +1,84 @@
+{
```
pzhokhov: I don't use VSCode so cannot really review / test this file. However, it seems like there are some run-specific arguments embedded into it. If we were to include it, I'd prefer a more generic version (maybe instead a doc describing how to generate this file)

pzhokhov:
@davidhopper2003 while it looks like a sensible change, could you look into fixing the failing travis tests for this PR? Looks like there a few and they seem to be related to the change. Please post here if you need some help with that.

pzhokhov:
closing this (upstream is changed to using method "spawn")

PR #751, opened 2018-12-04
[Update misc_util.py]: clean up unused helper functions

zuoxingdong:
A proposal for a clean-up, as it seems these two functions are rarely used.


pzhokhov:
the failing test suggests they are used after all :)

zuoxingdong:
@pzhokhov Right, `zipsame` was used for MPI related code, it is reverted back now.

PR #752, opened 2018-12-05
fix loading the demo file

shibei00:
There is a bug for loading the demo file. When using "np.load" without unpacking array from the dictionary, Numpy always loads the demo file and seeks the position for every visiting, which hurts efficiency. It results in loading the demo file slowly.


pzhokhov:
Confirmed. Let me not merge this though - I have fixed this bug in a bigger her refactor that's coming (currently in internal review), so once that is out, it should not be a problem anymore

PR #777, opened 2019-01-03
add an argument for importing extra modules from run

machinaut:


baselines/run.py
```diff
@@ -120,6 +120,11 @@ def build_env(args):
 
 
 def get_env_type(env_id):
+    # Re-parse the gym registry, since we could have new envs since last time.
```
pzhokhov: I think we can do the parsing only here; in fact, may make sense to move _game_envs to be a local variable in get_env_type

PR #787, opened 2019-01-13
Set frame_stack default to zero for retro

MatPoliquin:
frame_stack default needs to be zero otherwise in retro the frames get stacked twice

### DIFF
--- a/baselines/common/retro_wrappers.py
+++ b/baselines/common/retro_wrappers.py
@@ -221,13 +221,14 @@ def make_retro(*, game, state, max_episode_steps, **kwargs):
         env = TimeLimit(env, max_episode_steps=max_episode_steps)
     return env
 
-def wrap_deepmind_retro(env, scale=True, frame_stack=4):
+def wrap_deepmind_retro(env, scale=True, frame_stack=0):
     """
     Configure environment for retro games, using config similar to DeepMind-style Atari in wrap_deepmind
     """
     env = WarpFrame(env)
     env = ClipRewardEnv(env)
-    env = FrameStack(env, frame_stack)
+    if frame_stack > 0:
+        env = FrameStack(env, frame_stack)
     if scale:
         env = ScaledFloatFrame(env)
     return env


MatPoliquin:
Not sure how this error is related to my change

From the log:

=================================== FAILURES ===================================
________________________ test_discrete_identity[acktr] _________________________
alg = 'acktr'

    @pytest.mark.slow
    @pytest.mark.parametrize("alg", algos_disc)
    def test_discrete_identity(alg):
        '''
        Test if the algorithm (with an mlp policy)
        can learn an identity transformation (i.e. return observation as an action)
        '''
    
        kwargs = learn_kwargs[alg]
        kwargs.update(common_kwargs)
    
        learn_fn = lambda e: get_learn_function(alg)(env=e, **kwargs)
        env_fn = lambda: DiscreteIdentityEnv(10, episode_len=100)
>       simple_test(env_fn, learn_fn, 0.9)

baselines/common/tests/test_identity.py:40: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

env_fn = <function test_discrete_identity.<locals>.<lambda> at 0x7f58a151b598>
learn_fn = <function test_discrete_identity.<locals>.<lambda> at 0x7f58a150e730>
min_reward_fraction = 0.9, n_trials = 10000

    def simple_test(env_fn, learn_fn, min_reward_fraction, n_trials=N_TRIALS):
        np.random.seed(0)
        np_random.seed(0)
    
        env = DummyVecEnv([env_fn])
    
    
        with tf.Graph().as_default(), tf.Session(config=tf.ConfigProto(allow_soft_placement=True)).as_default():
            tf.set_random_seed(0)
    
            model = learn_fn(env)
    
            sum_rew = 0
            done = True
    
            for i in range(n_trials):
                if done:
                    obs = env.reset()
                    state = model.initial_state
    
                if state is not None:
                    a, v, state, _ = model.step(obs, S=state, M=[False])
                else:
                    a, v, _, _ = model.step(obs)
    
                obs, rew, done, _ = env.step(a)
                sum_rew += float(rew)
    
            print("Reward in {} trials is {}".format(n_trials, sum_rew))
            assert sum_rew > min_reward_fraction * n_trials, \
>               'sum of rewards {} is less than {} of the total number of trials {}'.format(sum_rew, min_reward_fraction, n_trials)
E           AssertionError: sum of rewards 8892.0 is less than 0.9 of the total number of trials 10000

baselines/common/tests/util.py:39: AssertionError
----------------------------- Captured stdout call -----------------------------

pzhokhov:
I don't think this  is a problem currently, actually, because we pass `frame_stack=1` as an argument to `wrap_deepmind_retro` (here: https://github.com/openai/baselines/blob/49a99c7d23bfa70a852efeb843a971d867815738/baselines/common/cmd_util.py#L79). While this does add the FrameStack wrapper, and then VecFrameStack wrapper on top of the vecenv, the inner FrameStack does not do anything. I suppose we can add logic to not add FrameStack wrapper of the size 1, but this should not change code behavior.

pzhokhov:
opening another PR on this: https://github.com/openai/baselines/pull/871 and closing current PR.

PR #789, opened 2019-01-15
replace save_variables / load_variables with save_state / load_state

zhaoyuxin8024:
With the pair of save_variables / load_variables, the q_value function
of the model loaded for testing is wrong. Changing from save_variables /
load_variables to save_state / load_state can solve this problem.

### DIFF
--- a/baselines/deepq/deepq.py
+++ b/baselines/deepq/deepq.py
@@ -7,7 +7,6 @@ import cloudpickle
 import numpy as np
 
 import baselines.common.tf_util as U
-from baselines.common.tf_util import load_variables, save_variables
 from baselines import logger
 from baselines.common.schedules import LinearSchedule
 from baselines.common import set_global_seeds
@@ -39,7 +38,7 @@ class ActWrapper(object):
                 f.write(model_data)
 
             zipfile.ZipFile(arc_path, 'r', zipfile.ZIP_DEFLATED).extractall(td)
-            load_variables(os.path.join(td, "model"))
+            U.load_state(os.path.join(td, "model"))
 
         return ActWrapper(act, act_params)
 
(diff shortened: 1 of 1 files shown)


pzhokhov:
could you provide a snippet of code / commands to reproduce? My understanding is that save_variables / load_variables should do the same thing as load_state / save_state.

zhaoyuxin8024:
It was found by experiments when I trained a new environment with DQN. I printed the q_values during training. The values when the training completed were totally different from those printed during testing. 

The q_value printed during training is obtained in deepq.py in the iteration:
`debug_func = debug["q_values"]
print("q_value of original state:", debug_func(list[None]), file=f)
`

What I used for save_variables / load_variables before in train_env.py：
`act, debug =deepq.learn( env,
        network=models.mlp(num_layers=1, num_hidden=64,activation=tf.nn.relu),
        lr=0.01,
        exploration_fraction=0.1,
        total_timesteps=5000,
        gamma=0.7,
        learning_starts=0,
)
act.save(model.pkl)
debug_func = debug["q_values"]
list1 = [[0,0,1,2,3,4]]
print(debug_func(list1))
`

In test_env.py:
`act, debug = deepq.learn(env, network='mlp', total_timesteps=0, load_path="model.pkl")
debug_func = debug["q_values"]
list1 = [[0,0,1,2,3,4]]
print(debug_func(list1))
`

I cannot find the detailed data now, but the behaviors of the two results are completely different. Hence, the model loaded for testing cannot select the correct action, even which can be realized perfectly at the later stage of training. But changing from save_variables / load_variables to save_state / load_state makes the q_values consistent. I'm also wondering why save_variables / load_variables cannot work for my environment.

PR #793, opened 2019-01-19
fix typo in assertion statement in subproc_vec_env.py

hsukyle:


pzhokhov:
fixed upstream (obs[0] does not actually have to be ordereddict there as we are returning dict), closing this

PR #800, opened 2019-01-24
release Internal changes

pzhokhov:
Release several internal changes to baselines - use `spawn` method in subprocvecenv and shmemvecenv, related changes to logger and the way mpi is imported.


pzhokhov:
@joschu note that I have changed the test_mpi_util logic a little bit in order to to a) unify it with other mpi tests, and b) fix the failure in which mpi is imported (and hence initialized) and then subprocess (inheriting the MPI-set env variables) calls mpirun.

PR #807, opened 2019-01-30
Removing * from tf.group

karanchahal:
The * sign is required in the function definition, not in the arguments. This made the code a little confusing. The answer of the usage of * is listed here: < https://stackoverflow.com/questions/400739/what-does-asterisk-mean-in-python >. Please let me know if I'm doing something wrong here. 

Cheers !

### DIFF
--- a/baselines/deepq/build_graph.py
+++ b/baselines/deepq/build_graph.py
@@ -427,7 +427,7 @@ def build_train(make_obs_ph, q_func, num_actions, optimizer, grad_norm_clipping=
         for var, var_target in zip(sorted(q_func_vars, key=lambda v: v.name),
                                    sorted(target_q_func_vars, key=lambda v: v.name)):
             update_target_expr.append(var_target.assign(var))
-        update_target_expr = tf.group(*update_target_expr)
+        update_target_expr = tf.group(update_target_expr)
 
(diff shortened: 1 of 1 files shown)


pzhokhov:
I disagree. Although it would work both ways (because tf.group actually flattens inputs: https://github.com/tensorflow/tensorflow/blob/6612da89516247503f03ef76e974b51a434fb52e/tensorflow/python/ops/control_flow_ops.py#L3685) - so it would work with lists, lists of lists etc), generally, when passing list of arguments into the function with *args, * is required on both sides. Consider the following:
```
def f(*args):
  for arg in args:
      print(arg)

f(1,2)   #  prints 1, newline, 2
f(*[1,2]) # same - prints 1, newline, 2
f([1,2]) # prints [1,2]
```
so in the code you are suggesting in the first argument to tf.group (that is, `inputs[0]` becomes the entire list update_target_expr). The documentation of tf.group (https://www.tensorflow.org/api_docs/python/tf/group) is somewhat vague as to what should inputs be, but to it sounds more like each input is an op / tensor, not a list of ops. Again, I tested both ways and they both work, but the current one seems more in-line with the docs.

pzhokhov:
as such, closing this

PR #817, opened 2019-02-05
Created a new PPO version with Random Network Distillation [WIP]

simoninithomas:
Hey there,

I implemented a new PPO version called `rnd_ppo` that use **Random Network Distillation as a model to generate intrinsic rewards.**

I tried to keep the code **as clean and clear as possible**.

This implementation still **suffers from exploding intrinsic rewards (and exploding rnd loss) as the training progress.**

# Some Explanations about the code
## RND Model
I build the RND model in a new file in /common folder called `RND.py`
We have 3 placeholders:
- `NEXT_STATE` (takes only one frame so size is 84,84,1)
- `RND_OBS_MEAN` : rnd observation normalization hyperparameters
- `RND_OBS_STD` : rnd observation normalization hyperparameters

We build the 2 networks:
- Prediction network,
- Target network

## Observation Normalization for RND
- First we instantiated a `RunningMeanStd` to keep track of obs mean and std in `RND.py`
- We perform initial normalization parameters by stepping a random agent in the environment for a small number of steps before beinning optimization in `Runner` with the function `init_obs_rnd_norm()`
- We update the observation normalization parameters in `Runner`.
- We perform the obs normalization in the `RND model` when we pass observations to the RND model.

## Calculate Intrinsic Rewards
- In order to calculate them, we call the function `calculate_intrinsic_rewards`.

## Intrinsic Reward normalization
- First we instantiated a `RunningMeanStd` to keep track of obs mean and std in `RND.py`
- We update the intrinsic reward normalization parameters in `Runner`.
-  We perform the intrinsic reward normalization in the `Runner`.

## Training RND
⚠️ Critical part,
In fact, in the @yburda implementation, I saw two things:
- We don't clip the value
- We train our ppo model and the predictor network of our rnd model with the same loss (total_loss).

I have still the problem of exploding intrinsic rewards, and I can't understand why it's doing that. I think that's related to the way I calculated the loss. Or a problem when updating the weights. I'm a little bit puzzled about where the problem comes from.

Have a nice day!

### DIFF
--- a/baselines/common/RND.py
+++ b/baselines/common/RND.py
@@ -1,0 +1,106 @@
+import tensorflow as tf
+import functools
+
+from baselines.common.tf_util import initialize
+
+from baselines.common.running_mean_std import RunningMeanStd
+
+from baselines.common.tf_util import get_session, save_variables, load_variables
+
+
+# Helpers functions
+def conv2d(inputs, filters, kernel_size, strides, activation):
+	return tf.layers.conv2d(inputs = inputs,
+								filters = filters,
+								kernel_size = kernel_size,
+								strides = strides,
+								activation = activation)
+
+def fc(inputs, units, activation):
+	return tf.layers.dense(inputs = inputs,
+								units = units,
+								activation = activation)
+
+class RND():	
+	def __init__(self, ob_space, proportion_of_exp_used_for_predictor_update):
+		self.sess = sess = get_session()
+
+		self.proportion_of_exp_used_for_predictor_update = proportion_of_exp_used_for_predictor_update
+
+		# CREATE THE PLACEHOLDERS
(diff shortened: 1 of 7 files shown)


gyunt:
hi, @simoninithomas 
I had read your code.
1) There are confusions of the order of `extrinsic` and `intrinsic` for input and output.  For example, `minibatch` in `runner.py`.
2) The output of a target network of RND should be gradient stopped. i.e, use `tf.stop_gradient()`.

Thanks,

ChuaCheowHuan:
Hi @simoninithomas , 

with respect to line 82 in RND.py,

```
# CALCULATE THE LOSS
# The loss is just the negative of IR
self.rnd_loss = -tf.reduce_mean(tf.square(self.target_next_feature - self.pred_next_feature))
```

could you please explain why `self.rnd_loss` is negative? Thank you.

Issue #802, opened 2019-01-25
observation normalization?


hankly:
https://github.com/openai/baselines/blob/b55eda1dde7292a56ad63d8a849198100bd26a17/baselines/gail/adversary.py#L69

In gail/adeversay.py Line 69.
I suppose it normalize obs.but ,is it something wrong?
 
it should be `obs = (obs_ph - self.obs_rms.mean) / self.obs_rms.std` instead?

mingfeisun:
https://github.com/openai/baselines/blob/5b41c926c7a852df3f0928afdf2429f96a3965cb/baselines/gail/adversary.py#L69
This line is inconsistent with:
https://github.com/openai/baselines/blob/5b41c926c7a852df3f0928afdf2429f96a3965cb/baselines/gail/mlp_policy.py#L36
So that line in **Adversary.py** should be:
`obs = (obs_ph - self.obs_rms.mean) / self.obs_rms.std`
And it makes more sense.

pzhokhov:
fixed in https://github.com/openai/baselines/pull/823

PR #819, opened 2019-02-08
minor normalization error in GAIL - #802 

sritee:
`obs = (obs_ph - self.obs_rms.mean / self.obs_rms.std)` corrected to 
`obs = (obs_ph - self.obs_rms.mean) / self.obs_rms.std`


pzhokhov:
fixed upstream, closing this

PR #823, opened 2019-02-15
fix bugs: obs_ph normalization in adversary.py

mingfeisun:
Fix bugs in GAIL, see https://github.com/openai/baselines/issues/802 for more details

### DIFF
--- a/baselines/gail/dataset/mujoco_dset.py
+++ b/baselines/gail/dataset/mujoco_dset.py
@@ -50,8 +50,12 @@ class Mujoco_Dset(object):
         # obs, acs: shape (N, L, ) + S where N = # episodes, L = episode length
         # and S is the environment observation/action space.
         # Flatten to (N * L, prod(S))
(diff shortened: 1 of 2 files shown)


baselines/gail/dataset/mujoco_dset.py
```diff
@@ -50,8 +50,12 @@ def __init__(self, expert_path, train_fraction=0.7, traj_limitation=-1, randomiz
         # obs, acs: shape (N, L, ) + S where N = # episodes, L = episode length
         # and S is the environment observation/action space.
         # Flatten to (N * L, prod(S))
-        self.obs = np.reshape(obs, [-1, np.prod(obs.shape[2:])])
-        self.acs = np.reshape(acs, [-1, np.prod(acs.shape[2:])])
+        if len(obs.shape[2:]) != 0:
```
pzhokhov: equivalent to len(obs.shape) > 2?

mingfeisun: Yes, you are right

PR #826, opened 2019-02-16
Replace np_random with np.random.RandomState()

amir-abdi:
Fixed bug within `baselines/common/tests/util.py` with regard to `np_random`.


pzhokhov:
fixed upstream (np_random is not actually used anymore), closing this

PR #847, opened 2019-03-11
ppo2: use time.perf_counter() instead of time.time() for time measurement

zlsh80826:
CLOCK_MONOTONIC(`time.perf_counter()`) is better than CLOCK_REALTIME(`time.time()`) when observing the duration of two events


pzhokhov:
lgtm

PR #849, opened 2019-03-11
fix freeze of ppo2

pzhokhov:


baselines/common/mpi_adam_optimizer.py
```diff
@@ -17,30 +17,27 @@ def compute_gradients(self, loss, var_list, **kwargs):
...
         avg_grads_and_vars = [(tf.reshape(g, v.shape), v)
                     for g, (_, v) in zip(avg_grads, grads_and_vars)]
         return avg_grads_and_vars
 
-def check_synced(sess, comm, tfstat):
+def check_synced(comm, localval):
     """
     Check that 'tfstat' evaluates to the same thing on every MPI worker
     """
-    localval = sess.run(tfstat)
```
pzhokhov: this is the offender - note that it is called in check_synced that is called in _collect_grads that is called within the graph

christopherhesse:
@pzhokhov do you know why it was freezing?

baselines/common/mpi_adam_optimizer.py
christopherhesse: So what caused this to stop working?

christopherhesse: Also the check_synced docstring should probably be updated.   Ideally it would also say exactly what kind of error this is supposed to cache.

pzhokhov:
I'll play with it for a little bit to see in which circumstances this is not working... a unit test here would be a good idea. Docstring and internal copy - agreed; as for which errors is this supposed to catch - frankly, I don't know in which circumstances these variables can go out of sync. I guess if we accidentally re-use them in some other portion of the graph?

joschu:
I added this because it's common to forget to initialize your variables to the same values, or (less commonly) if you update them in some other way than adam, to get them out of sync. It's probably bad to call sess.run inside sess.run, but it's not immediately obvious to me how to rewrite it in a different way.

christopherhesse:
Sounds good, want to update the docs @pzhokhov to include john's explanation and then see if it's reasonable to add a unit test?

pzhokhov:
yup! Also, found in which particular circumstances current version is crashing - there needs to be a session with a particular config (with inter_op_parallelism_threads == 1) - no clue why that particular config was crashing it, but there it is

joschu:
Ah, that actually makes some sense.

pzhokhov:
Ohhhh it totally does! It was not an infinite recursion rather a single-process deadlock - because only one thread was allowed to do sess.run, it was waiting for itself to complete the current run. Either way, let's remove nested sess.run logic (getting current session was also a little bit out of line with how we handle it elsewhere)

baselines/common/mpi_adam_optimizer.py
```diff
@@ -13,34 +18,61 @@ def compute_gradients(self, loss, var_list, **kwargs):
...
+    tf.set_random_seed(0)
+
+    a = tf.Variable(np.random.randn(3).astype('float32'))
+    b = tf.Variable(np.random.randn(2,5).astype('float32'))
+    loss = tf.reduce_sum(tf.square(a)) + tf.reduce_sum(tf.sin(b))
+
+    stepsize = 1e-2
+    # for some reason the session config with inter_op_parallelism_threads was causing
+    # nested sess.run calls to freeze
+    config = tf.ConfigProto(inter_op_parallelism_threads=2)
```
christopherhesse: Wait, does it freeze with threads=1 or threads=2?  If it freezes with threads=1 shouldn't this test use that?

pzhokhov: freezes with 1... my bad, forgot to revert this. Thanks!

PR #854, opened 2019-03-16
remove f-strings for python 3.5 compatibility

pzhokhov:


pzhokhov:
:) so do I, Chris, but it's either that or let python3.5 compatibility go

Issue #525, opened 2018-08-19
LSTM/RNN code is confusing


nahrak:
Hi! I couldn't run an LSTM policy with a custom DummyVecEnv, so I decided to try this:
_python run.py --env SeaquestNoFrameskip-v0 --network cnn_lstm --num_env 1_

Was getting the same error as with a custom env:
_AssertionError: Bad input for recurrent policy: batch size 32 smaller than nsteps 128_

Taking 1 minibatch works with a built-in environment:
_python run.py --env SeaquestNoFrameskip-v0 --network cnn_lstm --num_env 1 --nminibatches=1_

But this doesn't work with a custom DummyVecEnv:
_InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'ppo2_model/pi/Placeholder_2' with dtype float and shape [1]_

The documentation says to look in baselines.common/policies.py/lstm for documentation how to run recurrent policies, but there is nothing about recurrent nets or LSTMs in there.

pzhokhov:
Added extra docs. Closing this, feel free to reopen if some unclarity remains.

nahrak:
Hi!

I believe the number of minibatches should not be linked to the number of environments. Unless minibatches have a special meaning in PPO, a single environment can produce any number of minibatches, because they only serve for aggregating gradients. If PPO has 1024 steps, we may want to feed 16 batches of 64 examples each to the GPU or 32 batches of 32 examples, etc.

pzhokhov:
you are correct; however, in the current implementation the batch of data of the size nenv * nsteps is cannot be split in more than nenv minibatches when recurrent policy is used. In other words,  the slicing along the time axis is not implemented for recurrent policies (for non-recurrent ones the temporal structure does not matter, so it works ok as is). We'll fix that eventually, but it is not very high in priority. On the other hand, if you have a fix and a test, feel free to submit a PR :)

Issue #340, opened 2018-03-24
Is the lstm states evolution in ppo2.py correct?


Pengchengpcx:
In the ppo2.py, in function run() of Class Runner(), assuming that we are using lstm policy, the states from the output of this iteration will always be the input state of next iteration. Should we reset the states to be the initial states when the trajectory is done? i.e. in line 113, we should reset the lstm states when infos are done.

Issue #294, opened 2018-02-21
Is LSTM policy in PPO2 working correctly?


harry2636:
I tried on LSTM policy of PPO2 (with default hyper-parameter) in PongNoFrameSkipv4 environment but, it failed to improve the reward. It stuck at the minimum reward -21. I also tried on different nsteps value (4 and 1024) but it failed to.

However, when I  tried on CNN policy of PPO2 (also with default hyper-parameter), it succeeded to converge to the maximum reward of 21.

Is LSTM policy in PPO2 working correctly?

Guiadan:
worked perfect for me, on a different task. you can try to adjust network architecture to the task you're trying to solve. try changing the conv layers size rather than just changing the nsteps parameter.

harry2636:
@Guiadan Thanks! I will try it out.

davsuacar:
I tried on CNNPolicy in SonicTheHedgehog-Genesis and it works perfectly but when I tried to run LstmPolicy I got the following error:

```python

Traceback (most recent call last):
  File "/Users/david.suarez/Work/i+d/sonic-rl/.venv/lib/python3.6/site-packages/tensorflow/python/framework/ops.py", line 1567, in _create_c_op
    c_op = c_api.TF_FinishOperation(op_desc)
tensorflow.python.framework.errors_impl.InvalidArgumentError: Cannot reshape a tensor with 262144 elements to shape [0,4096,?] (0 elements) for 'model_1/Reshape_1' (op: 'Reshape') with input shapes: [512,512], [3] and with input tensors computed as partial shapes: input[1] = [0,4096,?].

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/david.suarez/Work/i+d/sonic-rl/ppo/ppo_agent.py", line 41, in <module>
    main()
  File "/Users/david.suarez/Work/i+d/sonic-rl/ppo/ppo_agent.py", line 36, in main
    total_timesteps=int(1e7)
  File "/Users/david.suarez/Work/i+d/sonic-rl/baselines/baselines/ppo2/ppo2.py", line 175, in learn
    model = make_model()
  File "/Users/david.suarez/Work/i+d/sonic-rl/baselines/baselines/ppo2/ppo2.py", line 170, in <lambda>
    max_grad_norm=max_grad_norm)
  File "/Users/david.suarez/Work/i+d/sonic-rl/baselines/baselines/ppo2/ppo2.py", line 18, in __init__
[...]

PR #859, opened 2019-03-21
RNN support for PPO2 

gyunt:
This is RNN support for PPO2 related to #294  #340 #525 issues.

- Improve the replay buffer management.
- RNN support for value/policy functions (i.e, `copy` options)
- Improve visualizations in Tesorboard graph.

I'm trying to make a minimal code changes. This, however, involves pretty many code changes originated from both the replay buffer and the RNNs implementation.

Because original replay buffers are individually created for each columns (ex. actions, observations, states, dones, ...), it is not easy to append a new column. The replay buffer have been refactored as dictionaries. Therefore, additional experiments could be easily implemented. In case of this PR, I had to refactor the replay buffer to save the memory state of RNN value/policy networks.

Original RNN code is confusing to me. When sampling a RNN memory state, the number of experienced memory states are not eqaul to number of steps. i.e, It was [1 x num_lstm]. I think original RNN in `baselines.common.models` are not designed for step by step sampling, but piling RNNs as usual in NLP or image processing. I had properly copied and modified the RNN code.

<del>When using a RNN model, it is quite difficult to manage the memory of RNNs. It had been moved to the inside of the model, and it will reduce the effort of managing it. For example ```actions, _, _, _ = model.step(obs, done)``` instead of ```actions, _, state, _ = model.step(obs ,S=state, M=done)``` </del>

Lastly, some `tf.variable_scope`s and `tf.name_scope` were added for Tensorboard graph visualization. It might be helpful for one's studying and debugging the code.

Thanks!

### Example
- `python -m baselines.run --alg=ppo2 --env=PongNoFrameskip-v4 `**`--network=ppo_cnn_lstm`** runs on an Atari Pong with `ppo_cnn_lstm` network.
- `python -m baselines.run --alg=ppo2 --env=Ant-v2 --num_timesteps=1e6 `**`--network=ppo_lstm --value_network=copy`** runs on a Mujoco Ant environment with `ppo_lstm` network whose value and policy networks are separated, but have same structure.

### PR vs Original 
The original version MLP and  PR version MLP/LSTM are almost same.
Original LSTM has poor performance.
image

### Networks of PR
PR version RNNs are also well performed when these are used with other FC layers.
image

### All Tested
image

### Params
Params
value_network: 'copy'  *except `original lstm` which only supports `shared`.*
nsteps: 128
num_env: 16
nminibatches: 16
noptepochs: 10
lr:   0.0003
num_timesteps: 2000000

_Other parameters are set to PPO2 default._


baselines/common/tests/util.py
```diff
@@ -17,20 +20,33 @@ def seeded_env_fn():
         tf.set_random_seed(0)
         model = learn_fn(env)
         sum_rew = 0
-        done = True
+        obs = [0]
+        done = [True]
+        args = inspect.getfullargspec(model.step).args
```
pzhokhov: is this variable being used somewhere?

baselines/ppo2/layers.py
```diff
@@ -0,0 +1,260 @@
+import numpy as np
+import tensorflow as tf
+
+from baselines.a2c.utils import ortho_init, fc
+from baselines.common.models import register
+
+
+@register("ppo_lstm")
+def ppo_lstm(nlstm=128, layer_norm=False):
```
pzhokhov: why are these new networks called `ppo_*`? They don't seem to be ppo-specific, I think they should be just as easily used with a2c, for instance?

baselines/ppo2/layers.py
```diff
@@ -0,0 +1,260 @@
...
+                            name='lstm_state',
+                            trainable=False,
+                            dtype=tf.float32,
+                            collections=[tf.GraphKeys.LOCAL_VARIABLES])
+
+        def _network_fn(input, mask, state):
+            mask = tf.to_float(mask)
+            initializer = ortho_init(np.sqrt(2))
+
+            h = tf.contrib.layers.conv2d(input,
```
pzhokhov: I think you can use baselines.common.nature_cnn here to create a stack of conv layers

baselines/ppo2/layers.py
pzhokhov: (and be moved into baselines.common.models?)

pzhokhov: oh I think I understand now, you are using different rnn functions and different way of managing state - only implemeneted in ppo - that's why these are ppo-specific for now. I have not read the rest of the PR yet, so more comments coming, but either way I'd like to have a singe view onto the RNN's in baselines (so if we are changing ppo, let's also change the rest of the algorithms compatible with rnns)

baselines/ppo2/layers.py
```diff
@@ -0,0 +1,260 @@
...
+    o = tf.nn.sigmoid(o)
+    u = tf.tanh(u)
+    c = f * c + i * u
+    h = o * tf.tanh(_ln(c, gc, bc))
+    s = tf.concat(axis=1, values=[c, h])
+
+    return h, s
+
+
+def gru(x, mask, state, nh, init_scale=-1.0):
```
pzhokhov: this is nitpicking, but I think gru should have the same call signature as lstm and lnlstm (and have `scope` argument)

baselines/ppo2/policies.py
```diff
@@ -0,0 +1,207 @@
...
+                value_latent = value_network_(encoded_x)
+
+        with tf.name_scope("save_rnn_memory"):
+            if policy_state or value_state:
+                next_states = tf.concat(next_states_list, axis=1)
+                state_info = {'current': state_placeholder,
+                              'next': next_states, }
+                update_op = []
+                index = 0
+                for state in [state for state in [policy_state, value_state] if state]:
```
pzhokhov: I think this can be expressed in a somewhat more readable (if maybe less concise) way as 
```
for state in [policy_state, value_state]:
    if state: 
        ...
```

pzhokhov: also, coming to think about it, even more readable version would be something like
```
for state, next_state in zip(state_list, next_state_list):
    if state and next_state:
       update_op.append(state.assign(next_state))
```
(in other words, I am not a fan of concatenating next_states and then slicing them again)

baselines/ppo2/policies.py
```diff
@@ -0,0 +1,207 @@
...
+                states=state_info,
+                sess=sess,
+                estimate_q=estimate_q,
+            )
+        return policy
+
+    return policy_fn
+
+
+def is_rnn_network(network):
```
pzhokhov: alright, this is hacky as f***. How about instead we a registration flag that indicates that policy is recurrent to the "register" decorator?

baselines/ppo2/policies.py
```diff
@@ -0,0 +1,207 @@
...
+                    assert callable(value_network)
+                    value_network_ = value_network
+
+                if is_rnn_network(value_network_):
+                    value_state, value_network_ = value_network_(encoded_x, dones)
+
+            if policy_state or value_state:
+                states_list = [state for state in [policy_state, value_state] if state]
+                states = tf.concat(states_list, axis=1)
+                state_placeholder = tf.placeholder_with_default(states, states.get_shape())
```
pzhokhov: If I understand correctly, this placeholder is not actually being returned anywhere outside of this function (it is stores in state_map, but then state_map is not being returned / referred to outside either). What is the point of making it a placeholder (i.e. how would external code feed data into it)?

pzhokhov:
@gyunt Thanks for putting so much work into this! Looks very interesting! 
If I understand correctly, there are really three big areas of changes (with 1) being the main focus):

1) make RNN state a tf variable and manage it internally in the model instead of externally in the runner
2) use dictionary for runner buffers
3) extra scopes

From my side, I have no problems with 2) and 3). As for 1) - the idea is appealing; the cases I can think of where one have to be careful with internal state management are
a) early resets - if we want to do an early reset, model.step(.., done=[1]) will have to be called. Arguably, external state management in that case is not the cleanest thing either. 
b) State reset to a previous state (I don't know how useful that really is though). If you add some way to pass in data into state_placeholder that will be doable.

Also putting @kcobbe in the loop as the main internal user of rnn policies with ppo2, and @joschu as the force behind ppo2 - they may have more insight/objections about this.

rkojcev:
@gyunt and @pzhokhov I really hope this PR moves forward. Looks really interesting.

gyunt:
Hi @pzhokhov, Thank you for your review.

I agree your point. I missed that only concatenation of the memory is needed to store the memory state easily. Therefore using local variables of Tensorflow graph are unnecessary, and it will not cause changing original model.step interface or tests. I will update the code soon.

baselines/ppo2/policies.py
gyunt: Unfortunately, `is_rnn_network` is a translation of the below code in `common.policies.build_policy`. 
To apply  the idea, refactoring of `@register` will be preceded.

```python baselines.common.build_policy
        with tf.variable_scope('pi', reuse=tf.AUTO_REUSE):
            policy_latent = policy_network(encoded_x)
            if isinstance(policy_latent, tuple):
                policy_latent, recurrent_tensors = policy_latent

                if recurrent_tensors is not None:
                    # recurrent architecture, need a few more steps
                    nenv = nbatch // nsteps
                    assert nenv > 0, 'Bad input for recurrent policy: batch size {} smaller than nsteps {}'.format(nbatch, nsteps)
                    policy_latent, recurrent_tensors = policy_network(encoded_x, nenv)
                    extra_tensors.update(recurrent_tensors)
```

baselines/ppo2/policies.py
gyunt: It will be removed.

baselines/ppo2/policies.py
gyunt: You're right. I going to apply the code.

baselines/ppo2/layers.py
gyunt: Okay 👍

baselines/ppo2/layers.py
gyunt: Okay 👍

baselines/ppo2/layers.py
gyunt: I'm not sure here, but probably I can use RNN networks in `common.models` without additional implementations.

baselines/common/tests/util.py
gyunt: I will be reverted.

gyunt:
> @gyunt and @pzhokhov I really hope this PR moves forward. Looks really interesting.

@rkojcev Thanks,

baselines/ppo2/policies.py
erniejunior: May I guide your attention to this related issue in the stable baselines: hill-a/stable-baselines#241? Here we discuss the same problem of how to flag a policy as recurrent.

baselines/common/tests/util.py
gyunt: reverted at 82ceb44

baselines/ppo2/layers.py
gyunt: replaced at 36aadd6

baselines/ppo2/layers.py
gyunt: GRU is removed at 36aadd6
I would like to make GRU implementation PR If it is necessary.

baselines/ppo2/layers.py
gyunt: @pzhokhov 
After I worked on it, I could use `baselines.a2c.utils.lstm` and `baselines.a2c.utils.lnlstm`. but it is unable to use RNNs in `baselines.common.models` because those declare `tf.placeholder`s.  

It is may arguable that using original `baselines.common.models` implementation and managing the memory of RNNs by another code. I had already implemented 'shared' version of it at in-house. It is complex enough. Because I had to manage not only initial_state, current_state, next_state of multiple RNNs but also name scope of copied network in this case.
(`policies.py` of the inhouse code: https://gist.github.com/gyunt/fd3ce10f2443a1ccc6d8253945904576  )

baselines/ppo2/policies.py
gyunt: removed at b6e6c52

baselines/ppo2/policies.py
gyunt: removed at b6e6c52

baselines/ppo2/policies.py
gyunt: Instead of `is_rnn_network()`, I refactored `@register` decorator.
(e6f0d98, b6e6c52)

gyunt:
Hi, @pzhokhov  
I applied what you reviewed and reduced code changes.

Thanks.

baselines/ppo2/policies.py
erniejunior: Cool!

EliasBarbaMoral:
Hi @gyunt,

Thank you very much for doing this! I have noticed that you have modified the code part where originally the experience is given in the gathered order [here](https://github.com/openai/baselines/blob/master/baselines/ppo2/ppo2.py#L158-L171), and just left the shuffled option [here](https://github.com/gyunt/baselines/blob/ppo2_rnn_pr/baselines/ppo2/ppo2.py#L170-L179).  Usually the power of LSTMs is in the ability to memorize important information from previous steps of a sequence, but by shuffling the experience, the hidden state losses meaning right? Is there some reasoning in doing that or I am missing something here?

gyunt:
@Elbarmo 
the replay buffer of the PR stores internal states to remember.

As far as I know, there is no assumption that supposes all time steps of RNNs should be trained at once.
Practically, storing all experienced internal states are not only expensive operations but tough to implement. Therefore, one may prefer to apply gradient descent methods to whole time steps of one example. 
In RL, we already have a replay buffer which could store internal state changes.

pzhokhov:
here's my thinking - I like the new version, and I want it to be available. Simultaneously, it makes working with ppo2 rather different than with a2c with recurrent policies, and has certain code duplication (mainly between ppo2/policies and common/policies). With that in mind, I propose we merge this PR in a separate branch (maybe `stateful_rnn`),  then refactor a2c to be able to use the same recurrent policies as ppo2, and ppo2/policies.py to be useable by the rest of policy gradient-like algorithms (trpo, acer, acktr, trpo_mpi). Once all of that is done, we can merge it into master. @gyunt what do you think?

gyunt:
Good. I would like to make the PR of one of those every week.

ZhuFengdaaa:
A meaningful and solid improvement. Can't wait to use.

pzhokhov:
okay, if everyone agrees, I am rebasing the commit to a new branch, `stateful_rnn`, and merging.

YueErro:
Hi @gyunt did you manage to run a trained network with your code? I have been trying this and I always get the following error:

```
model = model_fn(policy=policy, ob_space=ob_space, ac_space=ac_space, nbatch_act=nenvs, nbatch_train=nbatch_train,nsteps=defaults['nsteps'],ent_coef=defaults['ent_coef'],vf_coef=defaults['vf_coef'],max_grad_norm=0.5)
  File "/home/yue/github/ros2learn/algorithms/baselines/baselines/ppo2/model.py", line 76, in __init__
    value_clipped = VALUE_PREV + tf.clip_by_value(value - VALUE_PREV, -CLIPRANGE, CLIPRANGE)

```

Any tips or piece of code on how to load a pre-trained network and see the results?

Thanks

PR #866, opened 2019-04-02
fix: baselines.run --extra-import does not work with num_env>=2

afg984:
The import_module statement does not affect subprocess vec envs, they should be run inside subprocesses


pzhokhov:
extra imports are now supported directly through openai gym (that is, statement like `gym.make('roboschool:RoboschoolAnt-v1')` will first import roboschool, and then try to create RoboschoolAnt environment. As such, extra imports logic in baselines can be removed altogether (just removed it upstream :)), and I think this can be closed.

PR #867, opened 2019-04-03
Update dataset.py

kocabey:
The `Dataset` class shuffles the dataset even when `shuffle` is set to False. This creates issues for recurrent policies.

### DIFF
--- a/baselines/common/dataset.py
+++ b/baselines/common/dataset.py
@@ -7,7 +7,8 @@ class Dataset(object):
         self.enable_shuffle = shuffle
         self.n = next(iter(data_map.values())).shape[0]
         self._next_id = 0
-        self.shuffle()
+        if shuffle:
+            self.shuffle()
 
     def shuffle(self):
         if self.deterministic:


pzhokhov:
I think `shuffle=False` intended to disable shuffling every time the Dataset is fully read (but it is still shuffled once on creation), whereas `deterministic=True` flag disables shuffling altogether. So if you want fully repeatable Dataset, you can use it with `deterministic=True`. Also, I think out of all baselines algorithms only ppo1 uses it, and it is being deprecated; so I would recommend using ppo2 (which supports recurring policies too). Please post here if that does not quite address your issue, otherwise, I'll close this PR

kocabey:
In ppo1, it's used as:

`d = Dataset(dict(ob=ob, ac=ac, atarg=atarg, vtarg=tdlamret), shuffle=not pi.recurrent)`

Here it doesn't set `deterministic=True` for recurrent policies, so I thought that was not the intended behavior. Anyway thanks for the heads up on ppo2!

pzhokhov:
oh... yeah, that does look like a bug. Fixed upstream. That being said, test coverage in ppo1 is virtually non-existent, so there could be more bugs - better use ppo2 :)

PR #870, opened 2019-04-04
[Update vec_normalize.py] Avoid first reward to be very large

zuoxingdong:
I've tested in Mujoco environments, it seems current reward normalization results in the first reward to have very large magnitude and being clipped at +10/-10. Perhaps it is better to keep first reward unchanged and start normalization from the second one.

### DIFF
--- a/baselines/common/vec_env/vec_normalize.py
+++ b/baselines/common/vec_env/vec_normalize.py
@@ -25,7 +25,8 @@ class VecNormalize(VecEnvWrapper):
         obs = self._obfilt(obs)
         if self.ret_rms:
             self.ret_rms.update(self.ret)
-            rews = np.clip(rews / np.sqrt(self.ret_rms.var + self.epsilon), -self.cliprew, self.cliprew)
+            if self.ret_rms.count >= 2:
+                rews = np.clip(rews / np.sqrt(self.ret_rms.var + self.epsilon), -self.cliprew, self.cliprew)
         self.ret[news] = 0.
         return obs, rews, news, infos
 


pzhokhov:
While I see the problem here, I'd prefer a more statistically sound solution - for instance, an unbiased variance estimator in rms, maybe modify RunningMeanStd to optionally compute the unbiased variance, or initialize RunningMeanStd with a count 1 and variance 1.

zuoxingdong:
@pzhokhov If I got it correctly, it can be wrong estimation if we simply initialize count as 1. I've tried a numerically stable modification like following

```python
class RunningMeanStd(object):
    # https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm
    def __init__(self, epsilon=1e-4, shape=()):
        self.shape = shape
        self.mean = None#np.zeros(shape, 'float64')
        self.var = None#np.ones(shape, 'float64')
        self.count = None#1 + epsilon

    def update(self, x):
        x = np.asarray(x)
        assert x.ndim == len(self.shape) + 1
        if self.mean is None:
            self.mean = x[0]
            self.var = np.zeros(self.shape)
            self.count = 1
            
            x = x[1:]
        
        if x.size > 0:
            batch_mean = np.mean(x, axis=0)
            batch_var = np.var(x, axis=0)
            batch_count = x.shape[0]
            self.update_from_moments(batch_mean, batch_var, batch_count)
```

pzhokhov:
@zuoxingdong if you initialize variance and count to 1, it is not necessarily wrong, it just takes certain prior on variance into account (See, for instance, K. Murphy, "Machine Learning: A probabilistic perspective" (2012), section 4.6.2.2, p 134 - with nu0=1 and sigma0=1). This will lead to a smoother transition between first and second reward. 
That being said, dynamic observation and reward normalization is a crutch that should not ideally be needed (imagine sparse rewards setting, where rewards are zero for most steps - with or without prior you are likely to have variance collapse, and non-zero rewards will then be huge and clipped). The reason why dynamic normalization is there, to my understanding, is that noone went into the trouble of writing correct static normalization wrappers for mujoco environments (where observations and rewards can vary in large range). In many cases, roboschool analogs of mujoco environments have more regular observations and reward sequences, I would recommend using those.

PR #879, opened 2019-04-18
Add env_type: 'custom' which specifies worker id by passing it to SubprocVecEnv

showaykerker:
For custom environments using, for example, ROS, this gives the user an easy way to assign each worker into different topic.

### DIFF
--- a/baselines/common/cmd_util.py
+++ b/baselines/common/cmd_util.py
@@ -31,7 +31,7 @@ def make_vec_env(env_id, env_type, num_env, seed,
     mpi_rank = MPI.COMM_WORLD.Get_rank() if MPI else 0
     seed = seed + 10000 * mpi_rank if seed is not None else None
     logger_dir = logger.get_dir()
-    def make_thunk(rank):
+    def make_thunk(rank, **kwargs):
         return lambda: make_env(
             env_id=env_id,
             env_type=env_type,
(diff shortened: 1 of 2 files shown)


baselines/common/cmd_util.py
```diff
@@ -42,24 +42,30 @@ def make_thunk(rank):
...
-def make_env(env_id, env_type, mpi_rank=0, subrank=0, seed=None, reward_scale=1.0, gamestate=None, flatten_dict_observations=True, wrapper_kwargs=None, logger_dir=None):
+def make_env(env_id, env_type, mpi_rank=0, subrank=0, seed=None, reward_scale=1.0, gamestate=None, flatten_dict_observations=True, wrapper_kwargs=None, logger_dir=None, **kwargs):
     wrapper_kwargs = wrapper_kwargs or {}
     if env_type == 'atari':
         env = make_atari(env_id)
     elif env_type == 'retro':
         import retro
         gamestate = gamestate or retro.State.DEFAULT
         env = retro_wrappers.make_retro(game=env_id, max_episode_steps=10000, use_restricted_actions=retro.Actions.DISCRETE, state=gamestate)
+    elif env_type == 'custom':
```
pzhokhov: I think 'custom' and not atari or retro use case can be unified. Also, given that some of the kwargs are relevant to the wrapper, and some - to gym.make, I'd rather have an `env_kwargs` dictionary argument (instead of **kwargs)

baselines/run.py
```diff
@@ -112,7 +112,7 @@ def build_env(args):
         flatten_dict_observations = alg not in {'her'}
         env = make_vec_env(env_id, env_type, args.num_env or 1, seed, reward_scale=args.reward_scale, flatten_dict_observations=flatten_dict_observations)
 
-        if env_type == 'mujoco':
+        if env_type in ['mujoco', 'custom']:
```
pzhokhov: VecNormalize is rather specific to mujoco, are you sure your custom environment needs it?

pzhokhov:
all in all, this change is in a very good direction, but we have a very similar one (among other changes) from the internal repo lined up for merging: https://github.com/openai/baselines/pull/895 - please feel free to review / comment.

showaykerker:
I've checked the branch `internal`, that makes a lot more sense than mine.
I'll close the PR, thanks for your comment anyway!

PR #885, opened 2019-04-26
remove bullet extra, constrain gym version to be >= 0.10.0

pzhokhov:


setup.py
```diff
@@ -32,7 +29,7 @@
       packages=[package for package in find_packages()
                 if package.startswith('baselines')],
       install_requires=[
-          'gym',
+          'gym>=0.10.0',
```
christopherhesse: What about `gym~=0.10.0`?  Only difference is it would not automatically upgrade to `0.11.0` and potentially break.

PR #886, opened 2019-04-26
change rms 2 tfrms switch in vec_normalize to be more explicit

pzhokhov:


baselines/common/vec_env/vec_normalize.py
```diff
@@ -1,18 +1,22 @@
...
 
-
 class VecNormalize(VecEnvWrapper):
     """
     A vectorized wrapper that normalizes the observations
     and returns from an environment.
     """
 
     def __init__(self, venv, ob=True, ret=True, clipob=10., cliprew=10., gamma=0.99, epsilon=1e-8):
+        use_tf = False
```
christopherhesse: shouldn't this be a keyword argument so people can set it without changing the code?

baselines/common/vec_env/vec_normalize.py
```diff
@@ -1,18 +1,22 @@
...
         VecEnvWrapper.__init__(self, venv)
-        self.ob_rms = RunningMeanStd(shape=self.observation_space.shape) if ob else None
-        self.ret_rms = RunningMeanStd(shape=()) if ret else None
+        if use_tf:
+            self.ob_rms = TfRunningMeanStd(shape=self.observation_space.shape, scope='ob_rms') if ob else None
+            self.ret_rms = TfRunningMeanStd(shape=(), scope='ret_rms') if ret else None
+        else:
+            self.ob_rms = TfRunningMeanStd(shape=self.observation_space.shape) if ob else None
```
christopherhesse: I think this is supposed to be just `RunningMeanStd`

pzhokhov:
yes; although I think it will only be used in the ddpg smoke test (we don't actually have a separate VecNormalize unit test, although we should)

PR #891, opened 2019-05-03
Fix missing imports in test_monitor

amir-abdi:
The test_monitor function did not work due to missing imports.


pzhokhov:
similar fixes done in https://github.com/openai/baselines/pull/921, closing this

PR #895, opened 2019-05-03
release Internal changes

pzhokhov:


evisoft:
assert all(val==vals[0] for val in vals[1:]),\
            f'MpiAdamOptimizer detected that different workers have different weights: {vals}'

give me error in Python 3.5.2

christopherhesse:
Whoops, looks like we accidentally used an f-string, @pzhokhov?

Issue #843, opened 2019-03-06
HER MPI broadcasting issue with non-Reach environments


vitchyr:
The following command runs fine:

```
time mpirun -np 8 python -m baselines.run --num_env=2 --alg=her --env=FetchReach-v1 --num_timesteps=100000 
```

However, if I try changing the environment to the FetchPush-v1 or FetchPickAndPlace-v1, I get the following error:
When trying to run multiple MPI threads
```
Training...
Traceback (most recent call last):
  File "/home/vitchyr/anaconda2/envs/baselines2/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/vitchyr/anaconda2/envs/baselines2/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/vitchyr/git/baselines/baselines/run.py", line 246, in <module>
    main(sys.argv)
  File "/home/vitchyr/git/baselines/baselines/run.py", line 210, in main
    model, env = train(args, extra_args)
  File "/home/vitchyr/git/baselines/baselines/run.py", line 79, in train
    **alg_kwargs
  File "/home/vitchyr/git/baselines/baselines/her/her.py", line 181, in learn
    policy_save_interval=policy_save_interval, demo_file=demo_file)
  File "/home/vitchyr/git/baselines/baselines/her/her.py", line 59, in train
    logger.record_tabular(key, mpi_average(val))
  File "/home/vitchyr/git/baselines/baselines/her/her.py", line 20, in mpi_average
    return mpi_moments(np.array(value))[0]
  File "/home/vitchyr/git/baselines/baselines/common/mpi_moments.py", line 22, in mpi_moments
    mean, count = mpi_mean(x, axis=axis, comm=comm, keepdims=True)
  File "/home/vitchyr/git/baselines/baselines/common/mpi_moments.py", line 16, in mpi_mean
    comm.Allreduce(localsum, globalsum, op=MPI.SUM)
  File "mpi4py/MPI/Comm.pyx", line 714, in mpi4py.MPI.Comm.Allreduce
```
These different environments work for me if I run them without MPI.

I am using anaconda. My Python version is 3.6.2 and this is the output of `pip freeze`:
```
absl-py==0.7.0
astor==0.7.1
baselines==0.1.5
certifi==2018.11.29
cffi==1.12.2
chardet==3.0.4
Click==7.0
cloudpickle==0.8.0
Cython==0.29.6
[...]

PR #898, opened 2019-05-07
Issue 843

wecacuee:
This fixes issue #843


baselines/common/mpi_moments.py
```diff
@@ -9,7 +9,7 @@ def mpi_mean(x, axis=0, comm=None, keepdims=False):
     if comm is None: comm = MPI.COMM_WORLD
     xsum = x.sum(axis=axis, keepdims=keepdims)
     n = xsum.size
-    localsum = np.zeros(n+1, x.dtype)
+    localsum = np.zeros(n+1, np.float32) # x.dtype)
```
pzhokhov: is this change necessary? If I understand the logic correctly, you'll already have a correct dtype at this point (at least when being called from `her.mpi_average`)

PR #899, opened 2019-05-08
Add vision network from IMPALA/CoinRun paper

fiorenza2:
Adds the neural network that was used in the [IMPALA paper](https://arxiv.org/abs/1802.01561) and [CoinRun paper](https://arxiv.org/abs/1812.02341).

### DIFF
--- a/baselines/common/models.py
+++ b/baselines/common/models.py
@@ -26,6 +26,30 @@ def nature_cnn(unscaled_images, **conv_kwargs):
     h3 = conv_to_fc(h3)
     return activ(fc(h3, 'fc1', nh=512, init_scale=np.sqrt(2)))
 
+@register("impala_cnn")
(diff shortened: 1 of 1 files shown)


fiorenza2:
Apologies for any ignorance; not sure why flake8 is failing, and the line numbers seem to be incorrect.

pzhokhov:
flake8 complains because almost simultaneously with your PR we merged another one that already has  impala_cnn network :) Given that @kcobbe wrote that code, I am fairly certain it is the same network that was used for CoinRun paper; however, please review code here: https://github.com/openai/baselines/blob/9b68103b737ac46bc201dfb3121cfa5df2127e53/baselines/common/models.py#L112, and if it is the same with your PR, let's close it. If not, let's discuss how to unify the two.

fiorenza2:
No way! That's fair enough; had a look at the code, it's the same architecture, the only differences being minor:
* The way the weights are initialised in this PR are more consistent with the existing models
* This PR reuses the existing functions in `a2c/utils.py`, but should result in the same functionality

PR #907, opened 2019-05-18
Fix converting list of LazyFrames to ndarray

batist73:
Fix for issue #892. PR #875 changed behavior of \_\_getitem\_\_ method of [LazyFrames ](https://github.com/openai/baselines/blob/9b68103b737ac46bc201dfb3121cfa5df2127e53/baselines/common/atari_wrappers.py#L229) class. So, numpy wasn't able to convert list of LazyFrames instances in an [util method](https://github.com/openai/baselines/blob/9b68103b737ac46bc201dfb3121cfa5df2127e53/baselines/common/tf_util.py#L394).
I offer to reverse changes from #875, and introduce interface for getting count of frames and specific frames for LazyFrames instance.
Example of usage:
```python
import numpy as np

env = wrap_deepmind(make_atari('PongNoFrameskip-v4'), frame_stack=True)

state = env.reset()

print(type(state))
print(state.count())
print(state.frame(0).shape)

print(np.array([state]).shape)
```
The result:
```
<class 'baselines.common.atari_wrappers.LazyFrames'>
4
(84, 84)
(1, 84, 84, 4)
```

### DIFF
--- a/baselines/common/atari_wrappers.py
+++ b/baselines/common/atari_wrappers.py
@@ -254,6 +254,13 @@ class LazyFrames(object):
         return len(self._force())
 
     def __getitem__(self, i):
+        return self._force()[i]
+
+    def count(self):
+        frames = self._force()
+        return frames.shape[frames.ndim - 1]
+
+    def frame(self, i):
         return self._force()[..., i]
 
 def make_atari(env_id, max_episode_steps=None):


pzhokhov:
I am a bit sceptical about reversing #875 - it was a bug, and I think the PR was a good solution; also the snippet
```
import numpy as np

env = wrap_deepmind(make_atari('PongNoFrameskip-v4'), frame_stack=True)

state = env.reset()
print(np.array([state]).shape)
```
works for me (prints `(1, 84, 84, 4)`) on master branch (if I understand correctly, it does not work for you with the error similar to #892) with numpy 1.16.4. Could you specify the numpy version you are using so that I could replicate the issue? Thanks!

pzhokhov:
nvm, I can reproduce after all, looking into it

pzhokhov:
@batist73 okay, on a second thought, I agree - if there are aspects of the behavior different from numpy array, (such as indexing with one number that slices the last dimension instead of the first), it is better to have separate methods for that. Merging.

PR #908, opened 2019-05-21
Make SubprocVecEnv works with DummyVecEnv

iamhatesz:
This allows nested environments, e.g. for synchronous sampling (N parallel rollouts with M environments in series).


pzhokhov:
while I agree that N parallel rollouts each with M envs in series may be a good idea, I think this is better implemented through changing subprocvecenv directly - otherwise, you end up with multiple batch dimensions of the observations, actions and rewards which will likely break a lot of code in the algorithms... If modifying subprocvecenv to have more than one env per process does not quite cover your use case, please elaborate and provide examples :)

iamhatesz:
@pzhokhov you are right, it is much clearer to implement it directly in SubprocVecEnv, so I did it. Please consider this PR, if it is useful for others and worth merging.

baselines/common/vec_env/subproc_vec_env.py
```diff
@@ -4,51 +4,62 @@
...
         while True:
+            buf = [None for _ in range(nenvs)]
             cmd, data = remote.recv()
             if cmd == 'step':
-                ob, reward, done, info = env.step(data)
-                if done:
-                    ob = env.reset()
-                remote.send((ob, reward, done, info))
+                for i, env in enumerate(envs):
+                    buf[i] = list(env.step(data[i]))
```
pzhokhov: this looks a little ugly (and , and I doubt that it gives any noticeable performance boost. 
How about:
```python
for i, env in enumerate(envs):
  ob, reward, done, info = env.step(data[i])
  if done:
    ob = env.reset()
  buf[i] = (ob, reward, done, info)
remote.send(buf)
```
or even 
```python
def step_env(env, data):
   ob, reward, done, info = env.step(data)
   if done:
       ob = env.reset()
   return (ob, reward, done, info)
remote.send([step_fn(env, data_single) for env, data_single in zip(envs, data)])
```

baselines/common/vec_env/subproc_vec_env.py
```diff
@@ -4,51 +4,62 @@
...
-    def __init__(self, env_fns, spaces=None, context='spawn'):
+    def __init__(self, env_fns, spaces=None, context='spawn', in_series=1):
         """
         Arguments:
 
         env_fns: iterable of callables -  functions that create environments to run in subprocesses. Need to be cloud-pickleable
         """
         self.waiting = False
         self.closed = False
+        self.in_series = in_series
```
pzhokhov: please add the description of the new parameter into the docstring

baselines/common/vec_env/test_vec_env.py
```diff
@@ -67,6 +67,24 @@ def make_fn(seed):
...
 
+@pytest.mark.parametrize('dtype', ('uint8', 'float32'))
+def test_sync_sampling(dtype):
+    num_envs = 4
+    num_envs_in_series = 2
+    num_steps = 100
+    shape = (3, 8)
+
+    def make_fn(seed):
+        """
```
pzhokhov: this is a little shameful how in baselines we have docstrings for single-line functions like this, and no docstrings for monster functions with 15 parameters with obscure names :) Given that the test right above this code is the same way, I leave it to your discretion as to whether to keep the docstring or remove it. Let me put it like this - if the docstring is not there, I won't complain :)

baselines/common/vec_env/subproc_vec_env.py
```diff
@@ -4,51 +4,62 @@
...
-                remote.send(env.render(mode='rgb_array'))
+                for i, env in enumerate(envs):
+                    buf[i] = env.render(mode='rgb_array')
+                remote.send(buf)
             elif cmd == 'close':
                 remote.close()
                 break
             elif cmd == 'get_spaces_spec':
-                remote.send((env.observation_space, env.action_space, env.spec))
+                remote.send((envs[0].observation_space, envs[0].action_space, envs[0].spec))
```
pzhokhov: should this send the specs for all envs (rather than only the spec for the first env) ?

pzhokhov: oh nvm looks like we are only using the spec from the very first env anyways

iamhatesz:
@pzhokhov thanks for your input, I've addressed most of your comments

iamhatesz:
@pzhokhov do you think it can be merged?

pzhokhov:
reverted the setup.py TF check and merging.

PR #914, opened 2019-05-23
add tuple pdtype

JonathanRaiman:
Adding support for Tuple action spaces 😄

### DIFF
--- a/baselines/common/distributions.py
+++ b/baselines/common/distributions.py
@@ -275,6 +275,133 @@ class BernoulliPd(Pd):
     def fromflat(cls, flat):
         return cls(flat)
 
+
+
+def _np_cast(x, dtype):
+    """Numpy cast, equivalent to tf.cast"""
+    return x.astype(dtype)
+
+
+def decode_tuple_sample(pdtypes, x):
+    """
+    Cast and convert a sample from its dense concatenated state back to constituent parts.
+
+    Arguments
+    ---------
+
+    :param pdtypes: list<PdType>, a TuplePdType's child PdTypes.
+    :param x: np.ndarray or tf.Tensor.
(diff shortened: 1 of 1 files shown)


baselines/common/distributions.py
```diff
@@ -275,6 +275,133 @@ def sample(self):
...
+
+    def tuple_sample_concat(self, samples):
+        out = []
+        for sample, pdtype in zip(samples, self.pdtypes):
+            if len(pdtype.sample_shape()) == 0:
+                sample = tf.expand_dims(sample, axis=-1)
+            if sample.dtype != self.sample_dtype:
+                sample = tf.cast(sample, self.sample_dtype)
+            out.append(sample)
+        return tf.concat(out, axis=-1)
```
pzhokhov: If I read the logic correctly, this concatenation will not work if the sub-types have samples with different number of dimensions (or if the sizes of any dimensions but the last don't match). Should we add an assertion into the constructor that checks for that?

baselines/common/distributions.py
```diff
@@ -275,6 +275,133 @@ def sample(self):
...
+        return TuplePd
+
+    def pdfromflat(self, flat):
+        return TuplePd(self.sample_dtype(), self.internal_pdtypes, flat)
+
+    def param_shape(self):
+        return [sum([pdtype.param_shape()[0]
+                     for pdtype in self.internal_pdtypes])]
+
+    def sample_shape(self):
```
pzhokhov: this also looks like it won't work right for the sub-types with multidimensional samples

JonathanRaiman:
@pzhokhov Thanks! I'll work on addressing these points.

ronuchit:
@JonathanRaiman Just chiming in to say that this commit would be very helpful for my own research as well :) Could you describe the assumptions on the subspace dimensions that are being made by this code? Thank you!!

PR #917, opened 2019-05-30
add log_path flag to command line utility

lbertge:
as directed by the README, the command 
`python -m baselines.run --alg=ppo2 --env=PongNoFrameskip-v4 --num_timesteps=2e7 --save_path=~/models/pong_20M_ppo2
`
saves the learning curve logs by default to `/tmp/`, or otherwise set by the env variable `$OPENAI_LOGDIR`. This was not immediately obvious to me, so I thought it would improve the command-line usage to include a flag for specifying the log directory. Additionally, since both `save_path` and `load_path` rely on flags and not on env variables, I thought it would be best to keep `log_path` in line with that same functionality/behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ python -m baselines.run --alg=deepq --env=PongNoFrameskip-v4 --num_timesteps=1e6
 ```
 
 ## Saving, loading and visualizing models
+
+### Saving and loading the model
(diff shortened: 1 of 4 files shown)


lbertge:
OK. I updated the doc with an example, and included a note in case there were gotchas about its usage.

pzhokhov:
Awesome, thanks! Merging

PR #946, opened 2019-06-27
Remove model def from deepq.

tanzhenyu:


jmugan:
Did it move somewhere else? I don't see `cnn_to_mlp` anywhere.

Issue #953, opened 2019-07-03
Allow specifying an activation function for the output layer of the actor 


williamjshipman:
I wanted to test an architecture for PPO2 where the actor and critic share the hidden layers, but the actor's output layer has a `tanh` activation function instead of the default linear one. If I specify separate actor and critic networks then it is possible to change the actor's output layer to use `tanh`. However, I couldn't find an easy way to share the hidden layers and get the `tanh` activation. So, I went and added an extra activation function parameter to `PolicyWithValue.__init__`, `DiagGaussianPdType.pdfromlatent` and `_matching_fc`. This seems to be the easiest route, is this a good idea and is there any interest in adding this to Baselines? I'm happy to make the change for all the probability distribution classes.

DanielTakeshi:
@williamjshipman  I had to do something similar, and yeah, it was a bit of hacky solution and I put a new variable in `PolicyWithValue`. I tested and it seems correct. (It would be good to put an example of the source code changes here, though.)

williamjshipman:
@DanielTakeshi here's the [commit in my fork](https://github.com/openai/baselines/commit/ea09b04b14456df875e511366ba81e89191de583)

It seems like I can make the same change to every `pdfromlatent` function, but I was wondering why Baselines settled on always having a linear activation, in case there's a good reason.

Once done, I'm happy to make a PR to try and get it back into Baselines.

williamjshipman:
To clarify further, my own code calls `learn` in `ppo2.py` and passes the additional `activation` parameter. This gets captured by `**network_kwargs`, thus filtering through to `build_policy` without changes to `ppo2.py` or the other actor-critic methods.

PR #955, opened 2019-07-05
Allow actor network's output layer to have any activation function

williamjshipman:
At present, it isn't possible to easily create an actor network that has an activation function for its output layer while still sharing the hidden layers with the critic network. The only way is to have two separate networks. This PR fixes that, resolving #953  that I opened to discuss adding this feature.

### DIFF
--- a/baselines/common/distributions.py
+++ b/baselines/common/distributions.py
@@ -39,7 +39,7 @@ class PdType(object):
         raise NotImplementedError
     def pdfromflat(self, flat):
         return self.pdclass()(flat)
-    def pdfromlatent(self, latent_vector, init_scale, init_bias):
+    def pdfromlatent(self, latent_vector, init_scale, init_bias, activation=None):
         raise NotImplementedError
     def param_shape(self):
         raise NotImplementedError
@@ -61,8 +61,8 @@ class CategoricalPdType(PdType):
         self.ncat = ncat
     def pdclass(self):
         return CategoricalPd
-    def pdfromlatent(self, latent_vector, init_scale=1.0, init_bias=0.0):
-        pdparam = _matching_fc(latent_vector, 'pi', self.ncat, init_scale=init_scale, init_bias=init_bias)
+    def pdfromlatent(self, latent_vector, init_scale=1.0, init_bias=0.0, activation=None):
+        pdparam = _matching_fc(latent_vector, 'pi', self.ncat, init_scale=init_scale, init_bias=init_bias, activation=activation)
         return self.pdfromflat(pdparam), pdparam
 
     def param_shape(self):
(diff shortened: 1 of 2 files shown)


williamjshipman:
Hi community, is there anything I need to add to get this PR merged?

pzhokhov:
Thanks for the PR @williamjshipman! While I understand and agree with the problem statement, I think the solution in this PR is a little hacky. Here's my proposition (it is a little more involved in the code, however, I believe it will lead to a cleaner result):
1. allow `network_builder` functions return either tensor or a dict
2. dict can have keys such as `latent`, `policy_latent`, `policy_head`, `value_latent`, `value_head`, `recurrent_tensors`
3. modify `common.policies.build_policy` method such that `{policy|value}_head`  are used to output directly {policy|value} head, without adding any more dense layers; whereas `{policy|value}_latent` is used as input to a dense layers for {policy|value}. `latent` tensor is used as shared input to dense layers for policy and value outputs. 
5. When network_builder outputs single tensor, it is assumed to be equivalent to `{"latent": tensor}`
6. specify whether {policy|value} function is recurrent or not at the registration time (see registration logic at `common.models.register`); in the former case, network builder would also handle batch vs time dimensions and `recurrent_tensors` output. 

Again, I understand that's quite a handful of changes; however, I think this will lead to a noticeably cleaner outcome. Let me know if this suggestion makes sense and if you would like to work on it. Thanks!

williamjshipman:
Hi @pzhokhov, agreed about it being a hacky solution - it worked for my particular problem, but I wasn't using recurrent networks. I'm happy to tackle the changes. I'll post additional questions here as I go.

williamjshipman:
Hi @pzhokhov I've taken a first pass at a possible fix. Instead of separate `policy_network` and `value_network` functions, there's now one `network_builder` function passed to `common.policies.build_policy`. I've updated some of the networks in `common.models` to show how a `network_builder` would look. It returns the dictionary discussed above with policy and value function networks. So far so good if using `latent` or `{policy|value}_latent` fields in the dictionary. The `{policy|value}_head` fields are also there, but I need some advice on how the code should be restructured because the no. of outputs isn't available until after the latent (or head) tensors have been created. I can think of two routes:

1. Move the call to `make_pdtype` in `common.policies.PolicyWithValue.__init__` out to `common.policies.build_policy` so that we know how many actions there are before calling `network_builder`.
2. Leave the current order, forcing `network_builder` to find out from external code how many actions are needed, or just having to hard code the no. of actions in it.
3. Pass the environment to `network_builder`, allowing it to use its own logic to figure out the shape of the outputs. This feels like a bit of duplication of effort since this already happens in `make_pdtype`. If these two methods disagree on the no. of outputs, the current code adds an extra dense layer in the probability type.

williamjshipman:
That should fix the PPO2 test that was failing with CartPole-v0. DDPG and DQN errors also fixed.

The fix for PPO2 performance has me a bit confused though. All that was required was adding in the step to flatten the value function latent representation. While I see the reason for doing that, I don't see why it should affect PPO2 performance when `vf_latent` and `tf.layers.flatten(vf_latent)` have the same shape afterwards. I thought that all that flatten does is reshape the tensor, nothing else. Also, this change affects ACER performance in `test_cartpole.py`. Without flatten, it can achieve >199. With flatten that drops to around 133.

PR #958, opened 2019-07-08
Added a ThreadedVecEnv. 

RaphaelMarinier:
The ThreadedVecEnv allows parallelizing environments when subprocesses (SubprocVecEnv) cannot be used.
See the contents of the commit for details.


pzhokhov:
Thanks for the PR @RaphaelMarinier ! While I agree with your reasoning about when SubprocVecEnv will not work, I am still not convinced about the widespread utility of the ThreadedVecEnv (i.e. are there use cases in which we need embedded python interpreter to run parallel environments?). Simultaneously, the code is complex enough to have a relatively high debugging/maintenance burden. We can move it in a separate experimental branch; alternatively, please provide more justification of utility.

PR #970, opened 2019-07-29
entrypoint variable made public

NicoBach:
Hello!
With a recent commit in openai gym (openai/gym@dc91f43) the entry_point variable was changed from private to public, thus breaking baselines/baselines/run.py script.
Updating baselines/run.py in lines 35 and 129 from
env_type = env._entry_point.split(':')[0].split('.')[-1]
to
env_type = env.entry_point.split(':')[0].split('.')[-1]
fixes this problem. I hope, this can be useful.

Best regards,
Nico Bach

### DIFF
--- a/baselines/run.py
+++ b/baselines/run.py
@@ -32,7 +32,7 @@ except ImportError:
 _game_envs = defaultdict(set)
 for env in gym.envs.registry.all():
     # TODO: solve this with regexes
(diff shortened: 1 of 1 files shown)


guikarist:
I think the requirements for `gym` should also be changed to `0.14.0` at least.
https://github.com/openai/baselines/blob/0182fe1877e95b2ef0a82747c20bed1523fb5a3f/setup.py#L34

PR #976, opened 2019-08-04
Remove duplicate code in adaptive param noise.

tanzhenyu:


tanzhenyu:
@pzhokhov a small bug mentioned last time

PR #978, opened 2019-08-06
Baselines for Tensorflow 2.0.

tanzhenyu:
Please do note that:
1. ACER, ACKTR, GAIL is still under development by external
contributors.
2. HER is still under development by tanzheny@google.com.


README.md
```diff
@@ -150,7 +150,7 @@ respectively. Note that these results may be not on the latest version of the co
 To cite this repository in publications:
 
     @misc{baselines,
-      author = {Dhariwal, Prafulla and Hesse, Christopher and Klimov, Oleg and Nichol, Alex and Plappert, Matthias and Radford, Alec and Schulman, John and Sidor, Szymon and Wu, Yuhuai and Zhokhov, Peter},
+      author = {Dhariwal, Prafulla and Hesse, Christopher and Klimov, Oleg and Nichol, Alex and Plappert, Matthias and Radford, Alec and Schulman, John and Sidor, Szymon and Wu, Yuhuai and Zhokhov, Peter, and Tan, Zhenyu},
```
pzhokhov: @tanzhenyu I think these are in alphabetic order on last names, so you should move your name one up :)

baselines/bench/monitor.py
```diff
@@ -160,3 +162,27 @@ def load_results(dir):
     df['t'] -= min(header['t_start'] for header in headers)
     df.headers = headers # HACK to preserve backwards compatibility
     return df
+
+def test_monitor():
```
pzhokhov: given that the file name does not start with `test_` or end with `_test.py`, I think this test will not be found when running something like `pytest .` (for instance, travis will not actually trigger this test)

README.md
tanzhenyu: :-) Done.

baselines/bench/monitor.py
tanzhenyu: Ah I see you moved it to another folder :-)
Ok removing it for now...I shall add all missing tests in another pr.

baselines/a2c/utils.py
```diff
@@ -34,115 +18,20 @@ def _ortho_init(shape, dtype, partition_info=None):
...
+    with tf.name_scope(scope):
+        layer = tf.keras.layers.Conv2D(filters=nf, kernel_size=rf, strides=stride, padding=pad,
+                                       data_format=data_format, kernel_initializer=ortho_init(init_scale))
+    return layer
+
+def fc(input_shape, scope, nh, *, init_scale=1.0, init_bias=0.0):
+    with tf.name_scope(scope):
+        layer = tf.keras.layers.Dense(units=nh, kernel_initializer=ortho_init(init_scale),
+                                      bias_initializer=tf.keras.initializers.Constant(init_bias))
+        # layer = tf.keras.layers.Dense(units=nh, kernel_initializer=tf.keras.initializers.Constant(init_scale),
```
pzhokhov: the only difference between uncommented and commented code is ortho_init, right?

baselines/a2c/utils.py
```diff
@@ -152,131 +41,44 @@ def discount_with_dones(rewards, dones, gamma):
...
+        super(LinearTimeDecay, self).__init__()
+        self.initial_learning_rate = initial_learning_rate
+        self.name = name
+
+    def __call__(self, step):
+        with tf.name_scope(self.name):
+            initial_learning_rate = tf.convert_to_tensor(self.initial_learning_rate, name="initial_learning_rate")
+            dtype = initial_learning_rate.dtype
+            step_t = tf.cast(step, dtype)
+            return initial_learning_rate * step_t
```
pzhokhov: The name LinearTimeDecay seems a little confusing, given that the learning rate actually increases with time

baselines/bench/monitor.py
```diff
@@ -1,5 +1,6 @@
 __all__ = ['Monitor', 'get_monitor_files', 'load_results']
 
+import gym
```
pzhokhov: what's this import for?

baselines/common/cmd_util.py
```diff
@@ -187,7 +166,7 @@ def robotics_arg_parser():
 
 def parse_unknown_args(args):
     """
-    Parse arguments not consumed by arg parser into a dictionary
+    Parse arguments not consumed by arg parser into a dicitonary
```
pzhokhov: I like old spelling better :)

baselines/common/mpi_adam_optimizer.py
```diff
@@ -1,54 +1,46 @@
...
-        self.grad_clip = grad_clip
-        self.mpi_rank_weight = mpi_rank_weight
-        tf.train.AdamOptimizer.__init__(self, **kwargs)
-    def compute_gradients(self, loss, var_list, **kwargs):
-        grads_and_vars = tf.train.AdamOptimizer.compute_gradients(self, loss, var_list, **kwargs)
-        grads_and_vars = [(g, v) for g, v in grads_and_vars if g is not None]
-        flat_grad = tf.concat([tf.reshape(g, (-1,)) for g, v in grads_and_vars], axis=0) * self.mpi_rank_weight
-        shapes = [v.shape.as_list() for g, v in grads_and_vars]
-        sizes = [int(np.prod(s)) for s in shapes]
+        self.beta1 = 0.9
```
pzhokhov: what's the purpose of reimplementing adam update logic here (instead of inheriting it from tf.train.AdamOptimizer)?

baselines/common/vec_env/vec_env.py
```diff
@@ -145,7 +145,8 @@ class VecEnvWrapper(VecEnv):
 
     def __init__(self, venv, observation_space=None, action_space=None):
         self.venv = venv
-        super().__init__(num_envs=venv.num_envs,
+        VecEnv.__init__(self,
```
pzhokhov: these two are equivalent in python3, are they not?

baselines/run.py
```diff
@@ -10,14 +10,18 @@
...
 from baselines import logger
 from importlib import import_module
 
 try:
     from mpi4py import MPI
 except ImportError:
     MPI = None
+# Fix for loading Roboschool environments
+try:
+    from OpenGL import GLU
```
pzhokhov: why is this necessary? If anything, let's rather move it to roboschool package

baselines/a2c/utils.py
tanzhenyu: Yes forgot to clean it up (I was using it to prevent randomness and debugging)

baselines/a2c/utils.py
tanzhenyu: Yeah it is...I was using it for PPO, turns out we don't need this anymore. Removed.

baselines/bench/monitor.py
tanzhenyu: Removed.

baselines/common/cmd_util.py
tanzhenyu: Aha, turns out it is changed in a recent pr, but I didn't sync from it.
https://github.com/openai/baselines/pull/917/files

baselines/common/mpi_adam_optimizer.py
tanzhenyu: Great question -- I reimplemented so that I don't specifically inherit from tf Adam, the reason is that I need to wrap everything under tf.function, but MPI.Allreduce will only work with numpy, not symbolic tf tensors.

baselines/common/vec_env/vec_env.py
tanzhenyu: Yep. Changing it back.

baselines/run.py
tanzhenyu: You're right...should just be a fix on roboschool side.

tanzhenyu:
Thanks for the review! PTAL and merge :-)

tanzhenyu:
#https://github.com/tensorflow/tensorflow/issues/25349

PR #984, opened 2019-08-09
entry_point to _entry_point

Ujwal2910:
The code did not work and was not able to recognize entry_points but is able to resolve and train with _entry_points


pzhokhov:
please see comment in https://github.com/openai/baselines/commit/0182fe1877e95b2ef0a82747c20bed1523fb5a3f#commitcomment-34633147

PR #995, opened 2019-08-26
Add SRU in a2c.util

RonaldJEN:
Simple Recurrent Units for Highly Parallelizable Recurrence
[https://arxiv.org/abs/1709.02755](url)


pzhokhov:
Hi @RonaldJEN! Thanks for your PR! Unfortunately, at the moment we are not accepting new features to baselines (mostly due to lack of the resources to maintain existing codebase, let alone a bigger one). I would recommend re-submitting a PR to the stable-baselines project (https://github.com/hill-a/stable-baselines).

PR #1000, opened 2019-09-06
use allreduce instead of Allreduce 

pzhokhov:
(send pickled data instead of floats) - probably affects performance somewhat, but avoid element number mismatch. Fixes 998

### DIFF
--- a/baselines/common/mpi_moments.py
+++ b/baselines/common/mpi_moments.py
@@ -12,8 +12,9 @@ def mpi_mean(x, axis=0, comm=None, keepdims=False):
     localsum = np.zeros(n+1, x.dtype)
     localsum[:n] = xsum.ravel()
     localsum[n] = x.shape[axis]
-    globalsum = np.zeros_like(localsum)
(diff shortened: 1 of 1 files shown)


Ranxi:
@pzhokhov 
Hi, I ran into this problem again when use the `Normalizer` under the `her` directory for another NN customized by myself, which cooperate with DDPG to do learn policies.
```
  ....
  File "/xxx/normalizer.py", line 94, in synchronize
    local_sum[...] = self._mpi_average(local_sum)
  File "/xxx/normalizer.py", line 87, in _mpi_average
    MPI.COMM_WORLD.Allreduce(x, buf, op=MPI.SUM)
  File "mpi4py/MPI/Comm.pyx", line 714, in mpi4py.MPI.Comm.Allreduce
mpi4py.MPI.Exception: MPI_ERR_IN_STATUS: error code in status
```
After I modified the code in `normalizer.py:line 87` just as you did, the program ran into endless loop with two CPU cores in 100% usage.
What's the going on? Is there any thing I omit to use the normalizer?

PR #1022, opened 2019-10-28
Minor documentation, fix numpy warning at startup

sytelus:
Currently while running `train_pong.py` we get following numpy warning:

```
C:\ProgramData\Anaconda3\lib\site-packages\numpy\core\fromnumeric.py:3118: RuntimeWarning: Mean of empty slice.
```
This is because current code tries to do `episode_rewards[-101:-1]` on array of size 1. The simple fix only does that if size > 1.

Also, added few comments train_pong.py.

### DIFF
--- a/baselines/deepq/deepq.py
+++ b/baselines/deepq/deepq.py
@@ -306,8 +306,9 @@ def learn(env,
                 # Update target network periodically.
                 update_target()
 
-            mean_100ep_reward = round(np.mean(episode_rewards[-100:]), 1)
             num_episodes = len(episode_rewards)
+            mean_100ep_reward = round(np.mean(episode_rewards[-101:-1]), 1) if \
+                num_episodes > 1 else 0.0
             if done and print_freq is not None and len(episode_rewards) % print_freq == 0:
                 logger.record_tabular("steps", t)
                 logger.record_tabular("episodes", num_episodes)
--- a/baselines/deepq/deepq.py
+++ b/baselines/deepq/deepq.py
@@ -306,7 +306,7 @@ def learn(env,
                 # Update target network periodically.
                 update_target()
 
-            mean_100ep_reward = round(np.mean(episode_rewards[-101:-1]), 1)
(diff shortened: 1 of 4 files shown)


baselines/deepq/deepq.py
```diff
@@ -306,8 +306,9 @@ def make_obs_ph(name):
                 # Update target network periodically.
                 update_target()
 
-            mean_100ep_reward = round(np.mean(episode_rewards[-101:-1]), 1)
             num_episodes = len(episode_rewards)
+            mean_100ep_reward = round(np.mean(episode_rewards[-101:-1]), 1) if \
+                num_episodes > 1 else 0.0
```
pzhokhov: so in this case the first reward would always be 0? I would argue that NaN is a better value - for instance, in the case when all real rewards are negative

baselines/bench/monitor.py
```diff
@@ -56,26 +59,29 @@ def step(self, action):
...
     def update(self, ob, rew, done, info):
+        self.total_steps += 1
         self.rewards.append(rew)
         if done:
             self.needs_reset = True
             eprew = sum(self.rewards)
             eplen = len(self.rewards)
-            epinfo = {"r": round(eprew, 6), "l": eplen, "t": round(time.time() - self.tstart, 6)}
+            ep_t = time.time() - self.tstart
+            ep_dt = time.time() - self.episode_start
```
pzhokhov: dt is the episode duration, correct? Is is really necessary (given that it can be deduced from differentiating t)?

baselines/deepq/deepq.py
sytelus: I agree and make this change.

baselines/bench/monitor.py
sytelus: For the custom environment, I was working with `ep_dt` wasn't easily deducible from `ep_t` as there were few things happening before the next episode start. However, I understand this is probably not the case for most other environments so I can remove logging this one if needed.

PR #1027, opened 2019-11-04
GAIL-PPO

lbertge:
Implements PPO as the generator policy in GAIL. 
I noticed from the argparse methods, that PPO was supposed to be a choice, so I wondered if it was possible to support it. After some work, I have been able to get it to converge consistently, with similar performance characteristics as TRPO. For this reason, I believe it's a valuable addition to this repository. 

I've shifted some code around to make it a little less repetitive, but shouldn't change any method signatures or parameters. I did make some slight changes to the argparser, which in retrospect may be breaking, so I'm willing to go back and redo those.

The current GAIL codebase runs on TF 1.14. 

Initial Benchmarks:
image

image

GAIL-PPO has also been configured to support MPI. The graphs are not as good-looking, but I hope it helps support my case:
image

Some additional thoughts:
I have tried but have not had much success with integrating PPO2's implementation with GAIL. A bit more investigation led me to believe that PPO1 doesn't clip its [value loss](https://github.com/openai/baselines/blob/master/baselines/ppo1/pposgd_simple.py#L114), but PPO2 [seems to](https://github.com/openai/baselines/blob/master/baselines/ppo2/model.py#L75). I note some other suspected changes, but I'd like to know if this is worth investigating before opening up a new discussion. Thank you, I'd appreciate your time.

### DIFF
--- a/baselines/gail/run_mujoco.py
+++ b/baselines/gail/run_mujoco.py
@@ -61,33 +61,41 @@ def argsparser():
 
 
 def get_task_name(args):
-    task_name = args.algo + "_gail."
+    task_name = args.alg + "_gail."
     if args.pretrained:
         task_name += "with_pretrained."
     if args.traj_limitation != np.inf:
         task_name += "transition_limitation_%d." % args.traj_limitation
-    task_name += args.env_id.split("-")[0]
+    task_name += args.env.split("-")[0]
     task_name = task_name + ".g_step_" + str(args.g_step) + ".d_step_" + str(args.d_step) + \
         ".policy_entcoeff_" + str(args.policy_entcoeff) + ".adversary_entcoeff_" + str(args.adversary_entcoeff)
     task_name += ".seed_" + str(args.seed)
     return task_name
 
+def configure_logger(log_path, **kwargs):
+    if log_path is not None:
+        logger.configure(log_path)
+    else:
+        logger.configure(**kwargs)
 
 def main(args):
     U.make_session(num_cpu=1).__enter__()
     set_global_seeds(args.seed)
-    env = gym.make(args.env_id)
+
+    task_name = get_task_name(args)
+    args.checkpoint_dir = osp.join(args.checkpoint_dir, task_name)
+    args.log_dir = osp.join(args.log_dir, task_name)
+    configure_logger(args.log_dir)
+    gym.logger.setLevel(logging.WARN)
+
+    env = gym.make(args.env)
+    env = bench.Monitor(env, logger.get_dir() and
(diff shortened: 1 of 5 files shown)


pzhokhov:
Hi @lbertge Thanks for your PR, this looks like a neat addition! Unfortunately, at the moment we are not accepting new features to baselines (mostly due to lack of the resources to maintain existing codebase, let alone a bigger one). I would actually recommend re-submitting a PR to the stable-baselines project (https://github.com/hill-a/stable-baselines)

Kailiangdong:
Thank you. I am currently looking for baselines/gail using ppo. Finally , I'd like to have GAIL with ppo2 and wassersteinGAN.
Can you tell more about intergration of ppo2? I think it is worth. We can open a new discussion about that.      Thx.

baselines/gail/run_mujoco.py
```diff
@@ -129,7 +142,7 @@ def train(env, seed, policy_fn, reward_giver, dataset, algo,
         pretrained_weight = behavior_clone.learn(env, policy_fn, dataset,
                                                  max_iters=BC_max_iter)
 
-    if algo == 'trpo':
+    if alg == 'trpo':
```
Kailiangdong: Why you here still write ’trpo'? Thx

baselines/gail/run_mujoco.py
```diff
@@ -36,13 +41,13 @@ def argsparser():
...
     # Optimization Configuration
-    parser.add_argument('--g_step', help='number of steps to train policy in each epoch', type=int, default=3)
+    parser.add_argument('--g_step', help='number of steps to train policy in each epoch', type=int, default=1)
     parser.add_argument('--d_step', help='number of steps to train discriminator in each epoch', type=int, default=1)
     # Network Configuration (Using MLP Policy)
     parser.add_argument('--policy_hidden_size', type=int, default=100)
     parser.add_argument('--adversary_hidden_size', type=int, default=100)
     # Algorithms Configuration
-    parser.add_argument('--algo', type=str, choices=['trpo', 'ppo'], default='trpo')
+    parser.add_argument('--alg', type=str, choices=['trpo', 'ppo'], default='trpo')
```
Kailiangdong: default should be change to "ppo" right?

lbertge:
Hi all,

thank you for your responses. @pzhokhov, I will do my best to look at stable-baselines.

@Kailiangdong, you're welcome to open an issue on my fork if you want to discuss it further. You should be able to clone the repo and run it on TF 1.14.

Kailiangdong:
@lbertge  Thank you. It seems that your repository don't open the "Issue".
I have a question, why you update G only once in ppo. But the original one is to update G for three times.

PR #1066, opened 2020-01-10
support for tensorflow-cpu

xakiru:
https://pypi.org/project/tensorflow-cpu/


xakiru:
ops I'm using tf2 branch nvm

PR #1076, opened 2020-01-29
Monitor should close what it inherits

harry-uglow:
In my work I have used a monitored environment which has some resources it needs to close. Using Monitor was the problem as it doesn't call close() on its parent. This fixes it


chanel03:
harry-uglow:monitor_close

chanel03:
@harry-uglow

PR #1147, opened 2020-10-20
Align variables

jsrimr:
we call ```work_remotes, remote``` sequentially in line 80, but declared variable ```self.work_remotes, self.remotes = ..``` which is in reverse order.

This causes unnecessary confusion.


jsrimr:
need to exclude other commits

PR #1201, opened 2022-06-16
Update setup.py

dunnus:
Adding 'tensorflow-macos' for in list tf_pkg_name for the new tensorflow builds for apple silicon chips.

'distutils.LooseVersion' is depracated
replacing it with 'packaging.version.parse'


akshaygulabrao:
I was about to make a pull request for the exact same thing!!!

